PR #3, opened 2018-02-07
Add API endpoint for demo app

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3370)

Templates are no longer being served from cms, so I deleted them and their styles, hence the many files deleted.

image

image

## HAWK permission activated

image


config/urls.py
```diff
@@ -1,21 +1,20 @@
...
 
 urlpatterns = [
-    url(
-        r'^$',
-        core.views.HelloWorldView.as_view(),
-        name='hello-world'
-    ),
     url(
         r'^healthcheck/database/$',
         healthcheck.views.DatabaseAPIView.as_view(),
```
konradko: databaseapiview? Looks like leftover from repo init

richtier: it's there on purpose. its the database healthcheck api endpoint

PR #4, opened 2018-02-15
Add multi-language support

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3417)

translations

translations-api

I downgraded wagtail as `wagtail-modelstranslations` does not support wagtail 2.


stuaxo:
Hey @richtier I've managed to port a couple of wagtail libs to wagtail 2 - it turns out that for most things it's just the imports that moved, so you can port a lot of libs by changing this sort of thing:

```
from wagtail.wagtailadmin.blah import foo
```

To

```
try:
    from wagtail.admin.blah import foo
except ImportError:
    from wagtail.wagtailadmin.blah import foo
```

Which I got from some existing libraries that had ported.



Like this PR, which is a bit larger and includes some django-2 fixes

https://github.com/City-of-Helsinki/wagtail-svgmap/pull/14/files

(Note - they are happy with the PR, I just haven't worked out how to get travis to stop trying to test it with python 2.7, where it definitely won't work)


Or this PR that someone else did

https://github.com/moorinteractive/wagtail-bakery/pull/27/files

Basically, just a bunch of cutting and pasting and the upstream people seem to like it

PR #8, opened 2018-02-21
Add auto translate 

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3481)

auto-translations


core/templates/core/google-cloud-credentials.html
```diff
@@ -0,0 +1,12 @@
+{
```
richtier: google translate api client requires the credentials to be provided as a file. I added a command `generate_google_translate_cerdentials` to generate it on deployment.

konradko: could use

```
from django.template.loader import get_template_from_string

tpl = Template(get_template_from_string("My name is {{ my_name }}."))
```

core/models.py
```diff
@@ -50,17 +50,20 @@ def draft_url(self):
...
         path = reverse(
             'wagtailadmin_pages:add',
             args=(self._meta.app_label, self._meta.model_name, 2)
         )
         return urljoin(environment_url, path) + '?' + querystring
 
+    def serialize_fields(self):
+        return {
+            key: value for key, value in model_to_dict(self).items()
+            if value is not None
```
sdonk: You can use `filter()`

richtier: will do as a follow up change

PR #13, opened 2018-03-07
Add icon selection to lede

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3620)

image
image


core/helpers.py
```diff
@@ -22,24 +21,34 @@ def build_translated_fieldname(field_name):
...
-def make_language_panel(content_panels, language_code=None):
-    return [
-        FieldPanel(
-            build_localized_fieldname(content_panel.field_name, language_code),
-            classname=content_panel.classname
-        ) for content_panel in content_panels
-    ]
+def make_language_panel(untranslated_panel, language_code):
+    panel = copy.deepcopy(untranslated_panel)
+    if hasattr(panel, 'field_name'):
```
konradko: what is happening here?

core/helpers.py
```diff
@@ -22,24 +21,34 @@ def build_translated_fieldname(field_name):
...
+    return panel
 
 
-def make_language_panels(content_panels):
-    return [
-        ObjectList(make_language_panel(content_panels, code), heading=name)
-        for code, name in settings.LANGUAGES
-    ]
+def make_language_panels(content_panels, language_code=None):
+    translated_panel_builder = functools.partial(
```
konradko: this needs a docstring or some comments, not sure what is happening here

core/management/commands/upload_icons.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.core.management import BaseCommand
+from django.conf import settings
+from django.core.files.storage import default_storage
+
+from core import constants
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
```
konradko: why mgm cmd and not CMS user input?

core/tests/test_views.py
```diff
@@ -102,10 +102,14 @@ def test_add_page_prepopulate(
...
 
-    model_as_dict = model_to_dict(
-        translated_page,
-        exclude=['go_live_at', 'expire_at', 'case_study_image']
-    )
+    model_as_dict = model_to_dict(translated_page, exclude=[
+        'go_live_at',
+        'expire_at',
+        'case_study_image',
+        'lede_column_one_icon'
```
konradko: please check with a content designer that this name is familiar

core/wagtail_hooks.py
```diff
@@ -37,3 +39,11 @@ def translate_page(request, page):
     page.save_revision(user=request.user)
 
     return redirect(reverse('wagtailadmin_pages:edit', args=(page.pk,)))
+
+
+@hooks.register('insert_editor_css')
+def editor_css():
```
konradko: what does this do

makefile
```diff
@@ -166,7 +166,7 @@ DEBUG_SET_ENV_VARS := \
 	export AWS_SECRET_ACCESS_KEY=$$DIRECTORY_CMS_AWS_SECRET_ACCESS_KEY
 
 debug_webserver:
-	$(DEBUG_SET_ENV_VARS) && $(DJANGO_WEBSERVER)
+	$(DEBUG_SET_ENV_VARS) && $(UPLOAD_ICONS) && $(DJANGO_WEBSERVER)
```
konradko: why icons cannot be just part of static?

core/helpers.py
richtier: it's taking a list of English "panels" (basically, fields) and adding the french, german, russian, etc counterparts.

That is achieved by cloning the English panel and then changing the `field_name` of the clone.

Some panels are not fields, but really a fieldset - so we also loop over the fieldset's `children` and perform the above behaviour for the children too.

core/management/commands/upload_icons.py
richtier: uploading the icons via management command has no human interaction so is not subject to human error (the user might upload the wrong image, or call it the wrong name, or forget to upload one).

core/wagtail_hooks.py
richtier: [Add additional CSS files or snippets to the page editor.](http://docs.wagtail.io/en/v1.7/reference/hooks.html#insert-editor-css)

makefile
richtier: CMS static files (e.g., CSS files and javascript files needed for the CMS admin interface) are served by the CMS server*.

CMS media files (user-uploaded files that are needed to render the template on e.g., FAS) are served by S3.

the icons would be rendered by .e.g, FAS, so it makes sense to store them in S3 too, rather than served from the CMS server.

*or CloudFront, if the configuration is setup

core/tests/test_views.py
richtier: https://uktrade.atlassian.net/browse/ED-3690

core/management/commands/upload_icons.py
konradko: Don’t we want it to be flexible? Why a list instead of upload?

makefile
konradko: 🤙

core/helpers.py
konradko: Can you please add a comment

PR #30, opened 2018-04-04
Configure immutable file storage

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3906)

immutable


core/tests/test_storage_backends.py
```diff
@@ -0,0 +1,28 @@
...
+def test_delete_disabled(mock_delete):
+    backend = storage_backends.ImmutableS3Boto3Storage()
+
+    backend.delete('some/name.png')
+
+    assert mock_delete.call_count == 0
+
+
+def test_replace_disabled():
+    assert storage_backends.ImmutableS3Boto3Storage.file_overwrite is False
```
konradko: What will happen if you try to replace the file, error?

richtier: the storage backend checks if the name is already used, and appends a suffix if needed, thanks to `file_overwrite = False`

core/storage_backends.py
```diff
@@ -0,0 +1,17 @@
...
+
+    Extends:
+        S3Boto3Storage
+
+    """
+
+    file_overwrite = False
+
+    def delete(self, name):
+        pass
```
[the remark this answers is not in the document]
richtier: it's not an error. the flow. the "image" exists as two things:
 1) the `wagtai.wagtailImages.Image` model in the db, which has the url of the actual image on s3.
 2) the actual image on s3.

the user is allowed to delete `1`, but we want to prevent the storage backend from actually deleting `2`.

I'll add a comment to `delete` to explain this

core/tests/test_storage_backends.py
konradko: nice!

core/storage_backends.py
```diff
@@ -0,0 +1,33 @@
+from storages.backends.s3boto3 import S3Boto3Storage
+
+
+class ImmutableS3Boto3Storage(S3Boto3Storage):
```
konradko: The name is a bit misleading, an immutable store would be unchangeable, this one is just `undeletable`.

konradko: The `ImmutableS3Boto3Storage` name is for storage, not files. New files can be added to the bucket, thus it is not immutable. Like in Python, list it mutable, a tuple is not - bucket is like a list.

PR #31, opened 2018-04-04
Prevent attempting to copy related fields upstream

richtier:
Target environment will complain if we attempt to copy a related object that is not present e.g. an IndustryPage may be related to an Article with pk of 2 - but there is no guarantee that the target environment has an Article with pk of 2, and if it does there is no guarantee that Article 2 is the expected one.


richtier:
> Can we make the copying the related fields optional instead of preventing them?

The related field is just an integer primary keys of the related object's PK. 
 
There is no guarantee that an object with the same PK exists on the target environment.

If an object with the same PK exists on the target environment, there is no guarantee that the object is the expected object.

We could implement something to solve this, but I expect that would take a lot of dev effort.

richtier:
The use case this feature is supporting is no longer needed

PR #45, opened 2018-04-17
Show only three industries on home page

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3902)


find_a_supplier/models.py
```diff
@@ -466,7 +466,7 @@ def get_url_path_parts(self, *args, **kwargs):
         APIField('seo_title'),
         fields.APIIndustriesListField(
             'industries',
-            queryset=IndustryPage.objects.all()[0:9],
+            queryset=IndustryPage.objects.all()[:3],
```
konradko: feels like the magic number would better be in conf/env var setting with a default

PR #58, opened 2018-04-23
Enforce historic unique slugs

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-4126)


core/models.py
```diff
@@ -40,8 +46,32 @@ class Meta:
...
+        is_historically_unique = True
+        if page:
+            queryset = (
+                HistoricSlug.objects
+                .filter(slug=slug)
+                .exclude(page=page)
+                .only('page__title')
+            )
+            is_historically_unique = queryset.count() == 0
+        return is_currently_unique and is_historically_unique
```
richtier: results in this
image

core/tests/test_views.py
```diff
@@ -455,7 +455,20 @@ def test_lookup_by_slug(translated_page, admin_client):
...
+def test_lookup_by_slug_historic(translated_page, admin_client):
+    old_slug = translated_page.slug
+    new_slug = translated_page.slug = 'new-slug'
+    translated_page.save()
+    for slug in [old_slug, new_slug]:
+        url = reverse('lookup-by-slug', kwargs={'slug': slug})
+        response = admin_client.get(url)
+
+        assert response.status_code == 200
+        assert response.json()['id'] == translated_page.id
```
kowalcj0: @richtier should we also check if the old and new slugs lead to the same page?

core/models.py
```diff
@@ -40,8 +46,28 @@ class Meta:
...
         super().__init__(*args, **kwargs)
 
+    @staticmethod
+    def _slug_is_available(slug, parent, page=None):
+        is_currently_unique = Page._slug_is_available(slug, parent, page)
+        is_historically_unique = True
+        queryset = HistoricSlug.objects.filter(slug=slug).only('page__title')
+        if page:
+            queryset = queryset.exclude(page__pk=page.pk)
+        is_historically_unique = queryset.count() == 0
```
kowalcj0: This might be an overkill, but it'd be a bit more explicit if parentheses were put around the right hand part of assignment:
`is_historically_unique = (queryset.count() == 0)` as `=` & `==` can be hard to spot at first sight

PR #60, opened 2018-04-25
Allow marking industries as showcase

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3902)

image


find_a_supplier/models.py
```diff
@@ -603,7 +605,11 @@ def get_url_path_parts(self, *args, **kwargs):
         APIField('seo_title'),
         fields.APIIndustriesListField(
             'industries',
-            queryset=IndustryPage.objects.all().order_by('slug')[:3],
+            queryset=(
+                IndustryPage.objects
+                .filter(show_on_homepage=True)
+                .order_by('slug')[:3]
```
kowalcj0: @richtier Should this `magic number` be replaced with an env variable?

richtier: it's not a number that will change depending on environments.

The home page simply only wants to see three industries.

PR #63, opened 2018-04-25
Improve privacy and cookies page

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-4127)


core/models.py
```diff
@@ -202,6 +202,9 @@ def save(self, *args, **kwargs):
             self.slug = self.slug_identity
         super().save(*args, **kwargs)
 
+    def get_url_path_parts(self, *args, **kwargs):
```
konradko: what does this do?

export_readiness/models.py
```diff
@@ -71,6 +72,8 @@ class PrivacyAndCookiesPage(ExclusivePageMixin, BasePage):
         FieldPanel('title_en_gb'),
     ]
 
+    promote_panels = []
```
konradko: why is it important for this to be empty? Why is it not empty by default in `BasePage`?

PR #83, opened 2018-05-04
Select show case industries

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-4190)
[this task](https://uktrade.atlassian.net/browse/ED-4207)


core/models.py
```diff
@@ -191,6 +191,7 @@ class ImageHash(models.Model):
     @staticmethod
     def generate_content_hash(file):
         filehash = hashlib.md5()
+        file.open()
```
sdonk: Are you not closing the file?

richtier: when I do `with file.open('rb') as f` I get error telling me this is not a context manager.

I'll try manually calling `close()`

PR #91, opened 2018-05-23
Add get finance page, fix bug where creating app throws error

eadpearce:
https://uktrade.atlassian.net/browse/CMS-81

also fixes a typo in `APIBreadcrumsSerializer`


core/fields.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, name):
 
 class APIBreadcrumbsField(APIField):
     def __init__(self, name, app_label):
-        serializer = serializers.APIBreadcrumsSerializer(
```
richtier: crums :)

core/models.py
```diff
@@ -224,7 +224,7 @@ class Meta:
     @classmethod
     def allowed_subpage_models(cls):
         return [
-            model for model in super().allowed_subpage_models()
+            model for model in super().clean_parent_page_models()
```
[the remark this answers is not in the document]
eadpearce: inheriting from BasePage instead of Page ended up breaking this for some reason - `allowed_subpage_models` just returns an empty list. this seems to fix it

export_readiness/models.py
```diff
@@ -74,3 +80,90 @@ class PrivacyAndCookiesPage(ExclusivePageMixin, BasePage):
...
+        'wagtailimages.Image',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+
+    breadcrumbs_label = models.CharField(max_length=50)
+    banner_content = MarkdownField()
+    section1_content = MarkdownField()
```
richtier: for consistency with other places, could you use `section_one_content` `section_two_content` please

PR #96, opened 2018-06-04
Upgrade requirements

konradko:


requirements.txt
```diff
@@ -7,61 +7,59 @@
...
 draftjs-exporter==2.0.0   # via wagtail
-export-elements==0.8.1
-gevent==1.2.2
-google-api-core==0.1.4    # via google-cloud-core, google-cloud-translate
-google-auth==1.4.1        # via google-api-core
-google-cloud-core==0.28.0  # via google-cloud-translate
-google-cloud-translate==1.3.0
+export-elements==0.12.2
+gevent==1.3.2.post0
+google-api-core==1.2.0    # via google-cloud-core, google-cloud-translate
```
richtier: major version change `0.1.4` to `1.2.0`.we need to manually check if the functionality still works.

requirements.txt
```diff
@@ -7,61 +7,59 @@
...
 html5lib==0.999999999     # via bleach, wagtail
 idna==2.6                 # via requests
 jmespath==0.9.3           # via boto3, botocore
 markdown==2.6.11          # via wagtail-markdown
 mohawk==0.3.4             # via sigauth
-olefile==0.45.1           # via pillow
-pillow==4.3.0             # via wagtail
-protobuf==3.5.1           # via google-api-core, googleapis-common-protos
-psycopg2==2.7.3.2
+pillow==5.1.0             # via wagtail
```
richtier: major version change.

requirements.txt
konradko: we should test translations on every release

requirements.txt
konradko: e2e testing should pick up errors

PR #106, opened 2018-07-05
Don't reverse sort order for child pages

eadpearce:
https://uktrade.atlassian.net/browse/CMS-247


core/serializers.py
```diff
@@ -69,7 +69,7 @@ def to_representation(self, instance):
         # the subclass has. We are ordering on the application level
         sorted_data = sorted(
             data,
-            key=lambda x: x[self.sorting_key], reverse=True
+            key=lambda x: x[self.sorting_key], reverse=False
```
richtier: `reverse=False` is the default behaviour so it could be removed.

PR #107, opened 2018-07-09
Remove dev/staging from google

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-160)


conf/settings.py
```diff
@@ -51,7 +51,6 @@
     'django.contrib.messages',
     'raven.contrib.django.raven_compat',
     'django.contrib.sessions',
-    'config',
```
eadpearce: does this need to be 'conf' instead?

richtier: we never really needed this. the entries in `INSTALLED_APPS` help django collect static files, load models, locate templates, and locate template tags. As `conf/config` has none of those things we don't need it `INSTALLED_APPS`.

eadpearce: neat. the more u know

richtier: :rainbow:

PR #119, opened 2018-07-25
Remove streamfieds and unused content migration code

sdonk:


richtier:
:+1:

invest/migrations/0008_auto_20180725_1311.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('invest', '0007_auto_20180719_1414'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
richtier: I once inadvertently lost data by deleting the fields in the same release as the data migration and a edge case prevented the problem from being found in dev and stage.

what I did wrong:
1) add the new fields
2) data migration
3) delete the old fields
4) release to prod

what would have been better:
1) add new fields
2) data migration
3) release to prod
4) check if the new fields have the data in prod
5) delete the old fields
2) release to prod


this way we can avoid losing data if the data migration fails.

sdonk: We are doing the latter. The migrations have been applied to dev. Content will be pushed upstream to staging and production, so I need the db structure to be in place. There's no content on prod and staging atm. Content on dev has already been checked

PR #121, opened 2018-08-10
CMS-382 remove invest- from slugs

sdonk:


invest/management/commands/remove_invest_prefix_from_slugs.py
```diff
@@ -0,0 +1,17 @@
...
+from core.models import Service
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        pages = Page.objects.filter(
+                service__in=Service.objects.filter(name=cms.INVEST)
+            )
+        for page in pages:
```
konradko: seems like data migrations would be more suitable, as it is unlikely to ever be used again. But mgm command is nicely tested 👍

sdonk: Yeah, I did it as migration to begin with but it was untestable because of the way wagtail works. Even @richtier gave up

PR #125, opened 2018-08-22
Refactor lookup by slug

richtier:
[this task](https://uktrade.atlassian.net/browse/CMS-398)


core/migrations/0008_auto_20180809_1215.py
```diff
@@ -13,9 +13,9 @@ def add_service_to_pages(apps, schema_editor):
     Service = apps.get_model('core', 'Service')
     Page = apps.get_model('wagtailcore', 'Page')
     for page in Page.objects.all():
-        if hasattr(page, 'view_app'):
+        if hasattr(page, 'service_name_value'):
```
richtier: this is now a field rather than a property.

core/models.py
```diff
@@ -67,51 +79,35 @@ def __init__(self, *args, **kwargs):
         self.signer = signing.Signer()
         #  workaround modeltranslation patching Page.clean in an unpythonic way
         #  goo.gl/yYD4pw
-        self.clean = self._clean
+        self.clean = lambda: None
```
richtier: the checks in `_clean` are no longer needed - they are done as part of `save`

core/models.py
```diff
@@ -13,21 +14,26 @@
...
 from django.utils import translation
 from django.utils.text import mark_safe, slugify
-
 from core import constants, forms
 
 
 class HistoricSlug(models.Model):
-    slug = models.SlugField()
+    slug = models.SlugField(db_index=True)
+    service_name = models.CharField(
```
richtier: this field is duplicate of Page.service_name. Must do this because in unique_together we're unable to do ('slug', 'page__service_name') because Page does not contain service_name (yet another problem caused by Page not being abstract).

PR #128, opened 2018-08-22
Fix copy upstream

sdonk:


core/views.py
```diff
@@ -176,7 +178,14 @@ def get_form_class(self):
         return page_class.get_edit_handler().get_form_class()
 
     def get_parent(self):
-        return get_object_or_404(Page, id=self.kwargs['parent_pk']).specific
+        filter = filters.ServiceNameFilter(
```
richtier: `filter` is a buit-in object. please use another name :)

PR #130, opened 2018-08-23
Optimise lookup by slug

richtier:
Reduced number of database calls by 78%.

This change also facilitates caching server side as the server side page cache cannot also cache global breadcrumbs on a per-page basis.

before:

image

after:

image


conf/settings.py
```diff
@@ -93,7 +93,6 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
-    'wagtail.core.middleware.SiteMiddleware',
```
richtier: this hits the database, but we do not use multi-site features so we do not need it.

core/models.py
```diff
@@ -23,6 +27,21 @@
...
+        choices=choices.CMS_APP_CHOICES,
+        null=True,
+        db_index=True
+    )
+    label = models.CharField(max_length=50)
+    slug = models.SlugField()
+
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    page = GenericForeignKey('content_type', 'object_id')
```
richtier: generic relation allows our concrete pages to link to this model instead of doing page = ForeignKey(Page) ... page.specific thus reducing number of database hits when we later want to list all the pages with global breadcrumbs

core/views.py
```diff
@@ -60,11 +60,14 @@ class PageLookupBySlugAPIEndpoint(APIEndpointBase):
...
     detail_only_fields = ['id']
     filter_class = filters.ServiceNameFilter
+    authentication_classes = []
 
     def get_queryset(self):
         return Page.objects.all()
 
     def get_object(self):
+        if hasattr(self, 'object'):
+            return self.object
```
richtier: get_object was being called twice. now we cache get_object

core/migrations/0017_auto_20180823_1545.py
```diff
@@ -0,0 +1,58 @@
...
+                label_zh_hans=page.breadcrumbs_label_zh_hans,
+                label_fr=page.breadcrumbs_label_fr,
+                label_es=page.breadcrumbs_label_es,
+                label_pt=page.breadcrumbs_label_pt,
+                label_pt_br=page.breadcrumbs_label_pt_br,
+                label_ar=page.breadcrumbs_label_ar,
+            )
+
+
+def delete_bookmarks(apps, schema_editor):
```
kowalcj0: :+1:  for having reverse code in place

core/models.py
```diff
@@ -84,7 +103,7 @@ def save(self, *args, **kwargs):
                     'service_name': self.service_name,
                 }
             )
-        except IntegrityError:
+        except IntegrityError as error:
```
kowalcj0: the `error` alias is not required here

conf/settings.py
```diff
@@ -426,11 +425,21 @@ def base64_value(key):
...
 }
+
+if FEATURE_FLAGS['DEBUG_TOOLBAR_ON']:
+    INSTALLED_APPS += ['debug_toolbar']
+
+    MIDDLEWARE_CLASSES = (
+        ['debug_toolbar.middleware.DebugToolbarMiddleware'] +
+        MIDDLEWARE_CLASSES
+    )
+    INTERNAL_IPS = '127.0.0.1'
```
konradko: what's this env var?

conf/urls.py
```diff
@@ -66,3 +66,10 @@
     url(r'', include(wagtail_urls)),
 
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+
+
+if settings.FEATURE_FLAGS['DEBUG_TOOLBAR_ON']:
```
konradko: I'm surprised this works, had some problems with the order in which settings / urls.py are loaded before

conf/urls.py
```diff
@@ -66,3 +66,10 @@
...
+    import debug_toolbar
```
konradko: why not at the top :-)

conf/settings.py
richtier: it's used by django debug toolbar to turn on/off

conf/urls.py
richtier: took from the docs :)

PR #167, opened 2018-09-17
Add breadcrumbs to HPO page

richtier:


invest/models.py
```diff
@@ -937,7 +937,7 @@ class HighPotentialOpportunityFormPage(ExclusivePageMixin, BasePage):
 
     heading = models.CharField(max_length=255)
     sub_heading = models.CharField(max_length=255)
-
+    breadcrumbs_label = models.CharField(max_length=50)
```
konradko: why max 50?

richtier: we need some value. and all other breadcrumbs are 50 :)

PR #173, opened 2018-09-19
CMS-437 Lookup by full path

sdonk:


core/models.py
```diff
@@ -132,7 +132,17 @@ def get_url(self, is_draft=False, language_code=settings.LANGUAGE_CODE):
             querystring['lang'] = language_code
         if querystring:
             url += '?' + urlencode(querystring)
-        return url
+        return urlencode
```
richtier: this is the problem :)

PR #179, opened 2018-09-26
CMS-503 exread homepage

sdonk:


export_readiness/models.py
```diff
@@ -683,3 +681,39 @@ class ArticlePage(BasePage):
...
+        FieldPanel('title_en_gb'),
+        FieldPanel('slug'),
+    ]
+
+    api_fields = [
+        APIField('seo_title'),
+        APIField('news_title'),
+        APIMarkdownToHTMLField('news_description'),
+        APIArticleNewsPageListField('articles'),
+        APIMetaField('meta')
```
richtier: dont forget `seo_title` and `search_description` :)

they will be used for setting meta tags for google

PR #189, opened 2018-10-08
Server side cache

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-302)

_when do we invalidate cache?_. On page save, yes, we have nested external data in many pages though. e.g., Breadcrumbs of other pages. We need to avoid the breadrcumbs and other related data becoming stale.

A solution is to cache bust on related page save. Magic is avoided by requiring the dev to explicitly specify what external Pages each Page subscibes to.

We can also go beyond cache busting: we can prefetch the cache content too: on save we can delete the existing cache content with the newly updated content.

That introduced another problem: how do we ensure that the cache contains "accurate" data: if the cache is populated with data that is shaped differently from what the API endpoint normally returns then the cache content will be considered incorrect.

That problem is solved by calling the API endpoint when prefetching the cache content. Test client is used as that keeps the request internal to the app.

All this resulted in three classes:
 - `CacheDatabaseSubscriber`: listens for changes to the database and stores in the cache
 - `CachePopulator`: called by `CacheDatabaseSubscriber` to populate the cache on database changes
 - `PageCache`: called as a result of `CachePopulator` calling the API endpoints.

todo:
 * [x] test
 * [x] remove from cache on unpublish


core/views.py
```diff
@@ -37,7 +46,7 @@ def get_listing_default_fields(cls, model):
 
     @property
     def permission_classes(self):
-        permission_classes = [SignatureCheckPermission]
+        permission_classes = []#SignatureCheckPermission]
```
richtier: for testing. will not commit this :)

PR #198, opened 2018-10-18
Fix cache invalidation and population

richtier:
[this task](https://uktrade.atlassian.net/browse/CMS-553)


core/cache.py
```diff
@@ -131,15 +161,6 @@ def populate_many_if_live(cls, sender, instance, *args, **kwargs):
                 page_cache=cls.page_cache, instance=related_instance
             )
 
-    @classmethod
-    def delete_many(cls, sender, instance, *args, **kwargs):
```
richtier: this method was not very useful - calling `populate_many_if_live` instead adds more value.

PR #200, opened 2018-10-23
Add export readiness terms agreed field

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-648)


richtier:
change in solution

PR #202, opened 2018-10-24
CMS-560 Speed up test execution

sdonk:


makefile
```diff
@@ -8,7 +8,7 @@ test_requirements:
 	pip install -r requirements_test.txt
 
 FLAKE8 := flake8 . --exclude=migrations,.venv,node_modules
-PYTEST := pytest . -v --ignore=node_modules --cov=. --cov-config=.coveragerc --capture=no $(pytest_args)
+PYTEST := pytest . -v --ignore=venv --cov=. --cov-config=.coveragerc --capture=no $(pytest_args)
```
richtier: we can have multiple ignores `--ignore=venv --ignore=node_modules`

note I've experienced pytest trying to run python tests it (somehow?) found in `node_modules`

makefile
```diff
@@ -149,7 +149,7 @@ docker_build:
 DEBUG_SET_ENV_VARS := \
 	export PORT=8010; \
 	export SECRET_KEY=debug; \
-	export DEBUG=true ;\
+	export DEBUG=false;\
```
richtier: `DEBUG` is helpful in dev :)

PR #205, opened 2018-10-31
CMS-523 Make eu exit news slug an env var

sdonk:


richtier:
does `EU_EXIT_NEWS_LISTING_PAGE_SLUG` need adding to docker/env.json?

PR #215, opened 2018-11-23
Add marketing campaign page

eadpearce:
https://uktrade.atlassian.net/browse/CMS-625


core/views.py
```diff
@@ -36,7 +36,8 @@
     ex_read_models.HomePage: ex_read_serializers.HomePageSerializer,
     ex_read_models.ArticleListingPage: ex_read_serializers.ArticleListingPageSerializer,  # NOQA
     ex_read_models.TopicLandingPage: ex_read_serializers.TopicLandingPageSerializer,  # NOQA
-    ex_read_models.InternationalLandingPage: ex_read_serializers.InternationalLandingPageSerializer  # NOQA
+    ex_read_models.InternationalLandingPage: ex_read_serializers.InternationalLandingPageSerializer,  # NOQA
+    ex_read_models.CampaignPage: ex_read_serializers.CampaignPageSerializer
```
richtier: :smile:

export_readiness/apps.py
```diff
@@ -19,3 +19,4 @@ def ready(self):
         cache.EUExitInternationalFormPageSubscriber.subscribe()
         cache.EUExitDomesticFormPageSubscriber.subscribe()
         cache.EUExitFormSuccessPageSubscriber.subscribe()
+        cache.CampaignPageSubscriber.subscribe()
```
richtier: :muscle:

eadpearce: hell to the yes for caching 💯

export_readiness/serializers.py
```diff
@@ -151,3 +151,63 @@ def get_articles(self, object):
...
+    related_page_two_url = serializers.CharField(max_length=255)
+    related_page_two_heading = serializers.CharField(max_length=255)
+    related_page_two_description = serializers.CharField(max_length=255)
+    related_page_two_image = wagtail_fields.ImageRenditionField('fill-640x360')
+
+    related_page_three_url = serializers.CharField(max_length=255)
+    related_page_three_heading = serializers.CharField(max_length=255)
+    related_page_three_description = serializers.CharField(max_length=255)
+    related_page_three_image = wagtail_fields.ImageRenditionField(
+        'fill-640x360')
```
richtier: fancy

eadpearce: SUPER fancy! this makes sure that the images are the correct aspect ratio

export_readiness/models.py
```diff
@@ -464,6 +464,299 @@ def articles_count(self):
...
+class MarketingPages(ExclusivePageMixin, BasePage):
+    service_name_value = cms.EXPORT_READINESS
+    slug_identity = 'export-readiness-marketing-pages'
+
+    subpage_types = [
+        'export_readiness.CampaignPage',
+    ]
+
+    settings_panels = [
+        FieldPanel('title_en_gb'),
```
richtier: to make this even simpler we can do
```
    title_value = 'Marketing pages'
    settings_panels = []

        self.title = self.title_value
```

eadpearce: I was wondering how to do this ty!

eadpearce: does make the edit page a bit weird looking since there are no fields to fill in. people might think it's a bug. but since we only need to create these kind of pages/folders once that's not a big deal

PR #228, opened 2018-12-04
Add drf to all the remaining models

sdonk:


core/serializers.py
```diff
@@ -4,6 +4,7 @@
 
 
 class BasePageSerializer(serializers.Serializer):
+    id = serializers.IntegerField()
```
richtier: why expose the id? `slug` FTW :)

sdonk: I trusted this test https://github.com/uktrade/directory-cms/blob/master/core/tests/test_views.py#L300

Happy to delete it if we don't need to have it (I don't think we do)

PR #235, opened 2018-12-13
[DNM] Write translated pages to cache in single transaction

richtier:


core/cache.py
```diff
@@ -70,9 +63,47 @@ def delete(cls, slug, service_name, language_codes):
             ))
         cls.cache.delete_many(keys)
 
+    @classmethod
+    def transaction(cls):
```
richtier: this allows us to do:

```
with PageCache.transaction() as page_cache:
    page_cache.set(...)
    page_cache.set(...)
    page_cache.set(...)

```

and one a single write will occur - which means greater write performance for pages that are translated eight times therefore needing eight cache writes.

PR #236, opened 2018-12-13
fix page cache in dummy cache mode

richtier:
[DefaultCacheProxy](https://github.com/django/django/blob/master/django/core/cache/__init__.py#L98) does magic that results in `transaction` not being set if env is using DummyCache, resulting in error.

### DIFF
--- a/core/cache.py
+++ b/core/cache.py
@@ -7,7 +7,7 @@ from wagtail.core.signals import page_published, page_unpublished
 from wagtail.core.models import Page
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


core/cache.py
```diff
@@ -77,12 +77,14 @@ def __exit__(self, *args):
         return Transaction()
 
 
-class TransactionsactionalCache(DefaultCacheProxy):
```
richtier: I must have had severe head trauma when I wrote `TransactionsactionalCache` :rofl:

PR #237, opened 2018-12-14
Generate etag asynchronously

richtier:


core/cache.py
```diff
@@ -134,6 +134,8 @@ def populate_sync(cls, instance):
 
     @staticmethod
     def generate_etag(instance):
+        # This method can hit the database. It's slow. Don't call it
```
kowalcj0: handy comment!

PR #245, opened 2019-01-08
Inherit from common parent serializer

sdonk:


export_readiness/serializers.py
```diff
@@ -13,6 +13,23 @@
 )
 
 
+class RelatedPagesSerializer(BasePageSerializer):
```
eadpearce: can we call this something like `PageWithRelatedPagesSerializer` to avoid confusing it with `RelatedArticlePageSerializer`. or rename `RelatedArticlePageSerializer`. whichever works

sdonk: Valid point, I'll change to the first you suggested

export_readiness/serializers.py
```diff
@@ -13,6 +13,23 @@
 )
 
 
+class PageWithRelatedPagesSerializer(BasePageSerializer):
+    related_pages = serializers.SerializerMethodField()
+
+    def get_related_pages(self, object):
+        items = [
+            object.related_page_one,
```
richtier: `object` is a built in. can you use something like like `instance`, or `obj`, or `page` etc :)

sdonk: good point, done! (although it's in a protected scope)

richtier: ty

PR #251, opened 2019-01-14
CMS-670 Fix 500 bug on pages

sdonk:


core/views.py
```diff
@@ -318,3 +321,20 @@ def get_template_names(self):
...
+    def get(self, request, format=None):
+        data = {
+            'types':
+                list(
+                    map(
+                        lambda x: str(x.__name__),
+                        MODELS_SERIALIZERS_MAPPING.keys()
+                    )
+                )
+        }
```
richtier: list comprehension would be easier to read. will this work:
```
{
    'types': [item.__name__ for item in MODELS_SERIALIZERS_MAPPING.keys()]
}
```

PR #263, opened 2019-01-28
CMS-761 assign entry point to user

sdonk:


users/forms.py
```diff
@@ -0,0 +1,59 @@
...
+        help_text=_('Superusers have full access to manage any object '
+                    'or setting.'))
+
+    def __init__(self, user, *args, **kwargs):
+        """Creating user (session user) should only see the groups
+        that have the same tree entry point of their own group(s).
+
+              |
+             \||/
+              |/ <-- User group entrypoint, show only groups  # NOQA
```
sdonk: Flake8 thinks this is wrong Python...

PR #269, opened 2019-02-04
CMS-839 - Editors should be able to partially manage "<branch>" pages

kowalcj0:
New test cases that cover behaviour from the BDD scenario in this [ticket](https://uktrade.atlassian.net/browse/CMS-839):
- test_editors_can_create_child_pages
- test_editors_cant_create_child_pages_without_mandatory_data
- test_editors_cant_create_pages_in_branch_they_dont_manage
- test_editors_cannot_publish_child_pages
- test_editors_can_submit_changes_for_moderation
- test_editors_can_view_drafts
- test_editors_can_list_revisions
- test_editors_can_compare_changes_between_revisions


conftest.py
```diff
@@ -155,3 +163,106 @@ def django_db_setup(django_db_blocker):
...
+@pytest.fixture
+def branch_moderator_factory(root_page):
+    """Returns: listing page with a child page, editor group, user & user client
+
+    Moderators need: ['add', 'edit', 'publish'] permissions to perform
+    all required actions.
+    See:
+    http://docs.wagtail.io/en/v2.0/topics/permissions.html#page-permissions
+    """
+    class EditorFactory(object):
```
richtier: python3 :heart: 
```
    class EditorFactory:
````

File: conftest.py [hunk 2]
```diff
@@ -155,3 +163,106 @@ def django_db_setup(django_db_blocker):
...
+        ):
+            return DummyUserFactory.get(
+                root_page=root_page,
+                password=password,
+                permissions=permissions,
+                is_superuser=is_superuser,
+                is_staff=is_staff
+            )
+
+    return EditorFactory()
```
richtier: it would be easier to read if the class was defined outside of the fixture (with the `root_page` passed in during `get`)

richtier: the same comment for the other fixtures too

kowalcj0: that's a good idea. will implement it!

conftest.py
```diff
@@ -155,3 +163,106 @@ def django_db_setup(django_db_blocker):
 
         for connection in db.connections.all():
             connection.close()
+
+
+class DummyUserFactory(object):
```
sdonk: It's python 3, you don't need to inherit from object.

Can you not use a factoryboy factory class for consistency?

users/tests/test_permissions.py
```diff
@@ -0,0 +1,286 @@
...
+
+import export_readiness.tests.factories as exred_factories
+
+
+@pytest.mark.CMS_837
+@pytest.mark.django_db
+def test_branch_moderators_should_see_pages_only_from_their_branch(
+    branch_moderator_factory
+):
+    listing_1, article_1, _, _, client_1 = branch_moderator_factory.get()
```
sdonk: Instead of having losts of `_` can you return a named tuple and use dotted notation?

kowalcj0: what would be the benefit? not sure if it would make it more readable.

kowalcj0: ok, it's a bit more readable :+1: 
I changed it to `namedtuple` :)

conftest.py
kowalcj0: > It's python 3, you don't need to inherit from object.

Done

> Can you not use a factoryboy factory class for consistency?

Don't thinks so, as I'm basically reusing existing factories.
What model class would I need to use to define `class Meta:`?

PR #273, opened 2019-02-07
Add eu exit banner to exred landing page

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-1049)


sdonk:
You need to modify the factory and serializer with the new fields

conf/settings.py
```diff
@@ -71,6 +71,7 @@
     'wagtail.admin',
     'wagtail.core',
     'wagtail.api.v2',
+    'wagtail.sites',
```
richtier: root page uses a url name that uses this namespace.

sdonk: good catch

PR #275, opened 2019-02-08
CMD-837 editors and moderators should only see pages from their branch

kowalcj0:
New scenarios:
* test_branch_editors_should_only_see_pages_from_their_branch
* test_branch_moderators_should_only_see_pages_from_their_branch

*NOTE*
In admin Wagtail doesn't show to users. pages which don't belong to their *GroupPagePermission* but on the API Level it allows users to view pages from other branches!

I've also added 3 extra scenarios for CMS-840
- test_moderators_can_view_revisions_from_other_branches
- test_moderators_can_reject_revision
- test_moderators_cannot_reject_revision_from_other_branch

PR #280, opened 2019-02-12
CMS-836 Admins should be able to access all pages in any branch

kowalcj0:
This [ticket](https://uktrade.atlassian.net/browse/CMS-836)

### DIFF
--- a/users/tests/test_permissions.py
+++ b/users/tests/test_permissions.py
@@ -567,22 +567,22 @@ def test_admins_should_be_able_to_access_all_pages_in_any_branch(
     branch_2 = branch_editor_moderator_factory.get(branch_2_root_page)
 
     resp_1 = admin.client.get(
-        f'/admin/api/v2beta/pages/?child_of={branch_1_root_page.pk}&for_explorer=1'
+        f'/admin/api/v2beta/pages/?child_of={branch_1_root_page.pk}&for_explorer=1'  # NOQA
     )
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -214,7 +233,7 @@ def setup_pages_group_permission_user(
 
 
 class BranchEditorFactory:
```
sdonk: Same as the other PR. Please keep fixtures only in this file

conftest.py
```diff
@@ -245,3 +283,103 @@ def branch_editor_factory(root_page):
...
+        moderator_is_staff=False
+):
+    """Returns a listing page with a child page, user group, users & clients.
+
+    For Wagtail permission model check:
+    http://docs.wagtail.io/en/v2.0/topics/permissions.html#page-permissions
+    """
+    # ensure that only permissions supported by Wagtail are used
+    available_permissions = [p for p, _, _ in PAGE_PERMISSION_TYPES]
+    assert not (set(editor_permissions) - set(available_permissions))
```
sdonk: `ordered` and `==`?

conftest.py
kowalcj0: all 3 PRs are temporarily "daisy chained". Once this is addressed in #275 then it will be fixed here

conftest.py
kowalcj0: same answer as in #275 
this will produce a nice assertion error upon failure

PR #283, opened 2019-02-12
CMS-828 uk hq folder

sdonk:


great_international/cache.py
```diff
@@ -19,6 +19,11 @@ class InternationalMarketingPagesSubscriber(AbstractDatabaseCacheSubscriber):
     ]
 
 
+class InternationalUKHQPagesSubscriber(AbstractDatabaseCacheSubscriber):
```
eadpearce: the folder page is not requested by the client. does it need a cache subscriber?

PR #288, opened 2019-02-19
CMS-846 Tags and inheritance

sdonk:


great_international/models.py
```diff
@@ -76,26 +78,31 @@ class InternationalHomePage(ExclusivePageMixin, BasePage):
...
+        FieldPanel('tags', widget=CheckboxSelectMultiple)
+    ]
 
     def save(self, *args, **kwargs):
         self.title = self.get_verbose_name()
         return super().save(*args, **kwargs)
 
 
-class InternationalUKHQPages(ExclusivePageMixin, BasePage):
+class InternationalRegionPages(ExclusivePageMixin, BasePage):
```
sdonk: there are no live pages of this type

great_international/signals.py
```diff
@@ -0,0 +1,4 @@
+def inherit_tags_from_parent(sender, instance, *args, **kwargs):
+    parent = instance.get_parent().specific
+    if hasattr(parent, 'tags'):
+        instance.tags.add(*parent.tags.all())
```
richtier: could this be turned into a pre_save signal and do `bulk=True`? one db hit instead of two

PR #290, opened 2019-02-21
CMS-877 related pages on international homepage

sdonk:


great_international/serializers.py
```diff
@@ -1,42 +1,54 @@
...
             obj.related_page_three
         ]
-        serializer = RelatedArticlePageSerializer(
-            [item for item in items if item],
-            context=self.context,
-            many=True,
-        )
-        return serializer.data
+        pages = [item for item in items if item]
+        for related_page in pages:
```
richtier: line 20 and 21 are doing two loops but those could be consolidated into one:

```
for related_page in pages:
    if not related page:
        continue
    ...

File: great_international/serializers.py [hunk 2]
```diff
@@ -1,42 +1,54 @@
...
+            # currently only used for articles and campaigns
+            # assumes if it's not an article, it must be a campaign
+            if hasattr(related_page.specific, 'article_title'):
```
richtier: looks like the page's model class is being determined based on the presence of a particular field?

would something more direct like this be possible?
```
serializer_class = MODELS_SERIALIZERS_MAPPING[related_page.specific.__class__]
```

or 

```
if isinstance(related_page.specific.__class__, TheRelatedArticleClass):
    seriaizer_class = RelatedArticlePageSerializer
else:
    serializer_class = RelatedCampaignPageSerializer
serializer = serializer_class(related_page.specific)
```

because checking the presence of a field is like checking what to feed an animal based on if they have a wet nose: eventually we will be feeding dog food to a human with a cold :)

great_international/serializers.py
eadpearce: neat!

great_international/serializers.py
eadpearce: if it quacks like a duck though... (but i see your point. this makes more sense)

sdonk: I vote for `serializer_class = MODELS_SERIALIZERS_MAPPING[related_page.specific.__class__]` we use it in the view as well!

PR #293, opened 2019-02-25
CMS-894 Create international folders models

sdonk:


great_international/models.py
```diff
@@ -140,19 +141,34 @@ def save(self, *args, **kwargs):
...
+        'great_international.InternationalArticlePage'
+    ]
+
+    settings_panels = [
+        FieldPanel('title_en_gb'),
+        FieldPanel('slug'),
+    ]
+
     def save(self, *args, **kwargs):
-        self.title = self.get_verbose_name()
```
eadpearce: don't we still want to autogenerate the title here? a user may think they can edit the title to make the page into another country but the slug will stay the same

great_international/models.py
```diff
@@ -140,19 +141,34 @@ def save(self, *args, **kwargs):
...
+    subpage_types = [
+        'great_international.InternationalRegionalFolderPage'
+    ]
 
-    subpage_types = []
+    tags = ParentalManyToManyField(Tag, blank=True)
 
     settings_panels = [
+        FieldPanel('title_en_gb'),
+        FieldPanel('slug'),
```
eadpearce: can we use a predefined list of choices here?

PR #304, opened 2019-02-27
Cms 864 International Sector Page

Kenneth-Siu:


great_international/cache.py
```diff
@@ -3,6 +3,11 @@
 from great_international import models
 
 
+class InternationalSectorPageSubscriber(AbstractDatabaseCacheSubscriber):
+    model = models.InternationalSectorPage
+    subscriptions = []
```
richtier: the page relies on `InternationalArticlePage`, so please add `InternationalArticlePage` to the `subscriptions`

that makes the InternationalSectorPage cache update when the related pages update.

great_international/models.py
```diff
@@ -29,6 +29,301 @@ def get_required_translatable_fields(cls):
...
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+    section_two_heading_three = models.CharField(max_length=255)
+    section_two_body_three = models.CharField(max_length=255)
+
+    case_study_title = models.CharField(max_length=255)
+    case_study_description = models.CharField(max_length=255)
+    case_study_cta_text = models.CharField(max_length=255)
```
eadpearce: are we going to let people set the url for the CTA button in the CMS as well or is this hardcoded?

great_international/models.py
```diff
@@ -29,6 +29,301 @@ def get_required_translatable_fields(cls):
...
+        'wagtailimages.Image',
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+
+    section_three_heading = models.CharField(max_length=255)
+    section_three_teaser = models.CharField(max_length=255)
+
+    section_three_heading_one = models.CharField(max_length=255)
```
eadpearce: these field names are pretty similar. could we have `section_three_subsection_one_heading` instead? bit of a mouthful but it's more explicit

great_international/models.py
```diff
@@ -29,6 +29,301 @@ def get_required_translatable_fields(cls):
...
+    section_three_body_one = MarkdownField(blank=True, null=True)
+
+    section_three_heading_two = models.CharField(max_length=255)
+    section_three_teaser_two = models.CharField(max_length=255)
+    section_three_body_two = MarkdownField(blank=True, null=True)
+
+    next_steps_heading = models.CharField(max_length=255)
+    next_steps_description = models.CharField(max_length=255)
+
+    related_page_one = models.ForeignKey(
```
eadpearce: i think we want to be able to add articles and/or campaigns as related pages here (check with nas or michal to be sure). we've done this for the homepage already so you can copy the same code from there in the model and serializer

great_international/models.py
```diff
@@ -29,6 +29,301 @@ def get_required_translatable_fields(cls):
...
+
+    content_panels = [
+        MultiFieldPanel(
+            [
+                FieldPanel('heading'),
+                FieldPanel('sub_heading'),
+                ImageChooserPanel('hero_image'),
+                FieldPanel('heading_teaser')
+            ],
+            heading='Heading'
```
eadpearce: can you put the headings before the fields? just makes it a bit easier to read in the code

great_international/serializers.py
```diff
@@ -54,6 +54,48 @@ def get_related_pages(self, obj):
...
 
 
+class InternationalSectorPageSerializer(PageWithRelatedPagesSerializer):
+    heading = serializers.CharField(max_length=255)
+    sub_heading = serializers.CharField(max_length=255)
+    hero_image = wagtail_fields.ImageRenditionField('original')
+    heading_teaser = serializers.CharField(max_length=255)
+
+    section_one_body = serializers.CharField(max_length=255)
+    section_one_image = wagtail_fields.ImageRenditionField('original')
```
eadpearce: can you crop this one to the correct aspect ratio pls. instead of 'original' pass in 'fill-640x360' if it's landscape and 'fill-600x800' if it's portrait

great_international/migrations/0010_merge_20190227_1555.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('great_international', '0009_merge_20190225_1214'),
+        ('great_international', '0008_auto_20190227_1405'),
+    ]
+
+    operations = [
```
sdonk: Can you please create one migration only?

great_international/models.py
```diff
@@ -29,6 +29,339 @@ def get_required_translatable_fields(cls):
         return []
 
 
+class InternationalSectorPage(BasePage):
+    service_name_value = cms.GREAT_INTERNATIONAL
+    parent_page_types = []
```
sdonk: I assume you'll change this to be the SectorLandingPage when that's ready, an empty list will make the page unreachable in the admin

PR #306, opened 2019-02-28
CMS-906 Remove marketing page

sdonk:


great_international/tests/test_models.py
```diff
@@ -1,21 +1,48 @@
 import pytest
+from wagtail.core.models import Page
 
 from great_international import models
 from . import factories
 from export_readiness.tests import factories as exread_factories
 
 
-def test_app_models():
+def test_models_hierarchy():
```
sdonk: I will create the same test and structure in the other tests!

PR #312, opened 2019-03-04
Region aware caching

richtier:
[this task](https://uktrade.atlassian.net/browse/CMS-915)


great_international/cache.py
```diff
@@ -31,6 +33,7 @@ class InternationalCampaignPageSubscriber(AbstractDatabaseCacheSubscriber):
 class InternationalArticleListingPageSubscriber(
     AbstractDatabaseCacheSubscriber
 ):
+    cache_populator = RegionAwareCachePopulator
```
richtier: here's the trick to make region-aware caching work for the page

PR #314, opened 2019-03-04
CMS-864 Update fields for sector pages

Kenneth-Siu:
- Section 1 image has an optional caption field
- Smallprint in statistics is optional
- Teasers/body sections are text fields
- "Next Steps" section is hardcoded


great_international/models.py
```diff
@@ -248,7 +246,13 @@ class InternationalSectorPage(BasePage):
                 FieldRowPanel(
                     [
                         FieldPanel('section_one_body'),
-                        ImageChooserPanel('section_one_image')
+                        MultiFieldPanel(
```
eadpearce: 👍

PR #326, opened 2019-03-13
TOR-55: Add InternationalCuratedTopicLandingPage type

ababic:
@sdonk / @eadpearce

Here is the first of two new page types.

Although it's probably appropriate for this page type to use `ExclusivePageMixin`, I haven't here, because it involves hardcoding the slug on the Page type, and we don't yet know what that is going to be (We can always change this later if need be).

Please shout if any amends are needed :)


great_international/serializers.py
```diff
@@ -290,3 +290,55 @@ def get_child_pages(self, obj):
...
+    )
+    feature_four_url = serializers.CharField()
+
+    feature_five_heading = serializers.CharField()
+    feature_five_image = wagtail_fields.ImageRenditionField('original')
+    feature_five_image_thumbnail = wagtail_fields.ImageRenditionField(
+        'fill-640x360|jpegquality-60|format-jpeg',
+        source='feature_five_image',
+    )
+    feature_five_url = serializers.CharField()
```
sdonk: Can you please serialise the features as an array? There are dataproxy classes examples in the same file that you can look at

sdonk: Also, please add a test and a factory class for the new page type

ababic: @sdonk I think we're only doing this for related/child pages currently. I can't find any examples of taking collections of arbitrary fields and serialising them separately. I'll need to dig into the DRF docs to figure out the best way to do this, as it's not something I've ever had to do before.

core/tests/test_cache.py
```diff
@@ -245,6 +245,7 @@ def test_all_models_cached():
         export_readiness.models.AllContactPagesPage,
         great_international.models.InternationalRegionPage,
         great_international.models.InternationalLocalisedFolderPage,
+        great_international.models.InternationalCuratedTopicLandingPage,
```
eadpearce: we probably want to cache this page so please add it in great_international/cache.py

great_international/serializers.py
sdonk: Have a serializer like this

`class FeatureSerializer(serializers.Serializer):
    content = serializers.CharField()
    image = wagtail_fields.ImageRenditionField('original')
    image_thumbnail = wagtail_fields.ImageRenditionField(
        'fill-640x360|jpegquality-60|format-jpeg',
        source='feature_three_image',
    )
    url = serializers.CharField()
`

Have all the fields not mandatory and then use a dataproxy class like this one https://github.com/uktrade/directory-cms/blob/develop/great_international/serializers.py#L68


https://github.com/uktrade/directory-cms/blob/develop/great_international/serializers.py#L68

ababic:
@eadpearce @sdonk Thank you for the review both :) I'll shout out again when these bits are sorted.

ababic:
@sdonk, All changes implemented now, as far as I can see. Please shout if you spot anything amiss.

And apologies for the flake8 errors... have just switched over to MS Code and am still working on the editor configuration :)

PR #332, opened 2019-03-18
Set up guide added to invest home page with migrations and all tests …

jessgilbert:
image
image

Replaced set up guide section on invest home page with link to set up guide page


invest/serializers.py
```diff
@@ -188,6 +188,10 @@ class InvestHomePageSerializer(BasePageSerializer):
     sector_title = serializers.CharField(max_length=255)
     sector_button_text = serializers.CharField(max_length=255)
     setup_guide_title = serializers.CharField(max_length=255)
+    setup_guide_content = core_fields.MarkdownToHTMLField()
```
erdem: Is there any unit test case needs to be updated for these fields? The fields need to add expected fields.

jessgilbert: Yes in "invest/tests/factories.py" I have added the extra required field so all tests pass

MichaelWalker: I don't think so sadly - it doesn't look like there are currently any unit tests for that serializer (unless I'm missing something). The serializer tests for invest only seem to test the HPO form page at the moment. 

We are about to make a larger set of changes to this page though (and therefore to this serializer) so I suggest we leave it for this commit, and add a new unit test to cover the class when we make those changes (later today)

MichaelWalker: Actually - having taken a closer look, I'm slightly confused.

From the codebase, there seems to be
- test_serializers.py: which appears to test that the meta-class magic works as expected where we've done thing more than just simple fields (which are effectively covered by the base classes)
- test_views.py: which appear to be more end-to-end tests.

sdonk: We test some of that field functionality in here https://github.com/uktrade/directory-cms/blob/develop/core/tests/test_helpers.py#L54

PR #343, opened 2019-03-20
change format of serializer

WillTaylor22:


activitystream/serializers.py
```diff
@@ -14,7 +14,7 @@ def to_representation(self, obj):
             'type': 'Update',
             'published': obj.last_published_at.isoformat('T'),
             'object': {
-                'type': ['Document', 'dit:cms:Article'],
+                'type': 'Article',
```
richtier: I get the impression this conf informs AS how to serialize/deserialize the "record" - is `Article` not too generic - the `dit:cms:` seems very helpful for specificity

WillTaylor22: Nah it's completely arbitrary information/styling, AS doesn't parse anything, nor is it really required by the W3 ActivityStream style guide.

PR #353, opened 2019-03-25
Integrate link chooser with SimpleMDE, using slug links (TOR-83, CMS-991)

thibaudcolas:
Addresses [TOR-83](https://uktrade.atlassian.net/browse/TOR-83), [CMS-991](https://uktrade.atlassian.net/browse/CMS-991) (same task).

This integrates Wagtail’s existing link chooser UI with the current Markdown editor, so end users can create internal links without having to manually enter the slug of pages. Links are created with the same syntax as before, so for Markdown processing this should work the same. And for users who want to manually enter links, this is also doable like before.

Quick GIF demo:

tor-83-demo

I wasn't sure how to test this, considering all of the behavior is in client-side JS, so here's the manual test plan I followed:

- Features
  - [x] Internal page links on selected text should decorate the text
  - [x] Internal page links without a selection should have empty link brackets (`[]`)
  - [x] External/email links with link text override, and selected text, should override
  - [x] External/email links with link text override, without a selection, should insert the link text
- Cross-browser tests:
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11
  - [x] Edge
- Other
  - [x] Supports multiple Markdown editors on the page (kind of obvious – but my previous PoC didn't! https://github.com/torchbox/dit_directory_cms_poc/pull/1)
  - [x] No regression in the operation of the image control in the toolbar (because of `window.prompt` override)
  - [x] Error message if the link chooser modal does not load

### DIFF
--- a/core/static/core/js/refresh_codemirror.js
+++ b/core/static/core/js/refresh_codemirror.js
@@ -38,17 +38,16 @@ function _replaceSelection(cm, active, startEnd, url) {
     }, 1);
 }
 
-const windowPrompt = window.prompt;
+var windowPrompt = window.prompt;
+window.prompt = windowPromptOverride;
 
-const OVERRIDE_TEXT = 'windowPromptOverride';
+function windowPromptOverride(value) {
+    var isOverride = value instanceof SimpleMDE;
 
-const windowPromptOverride = (callback, text) => {
-    const shouldOverride = text === OVERRIDE_TEXT;
-
-    if (shouldOverride) {
-        callback();
+    if (isOverride) {
+        openLinkChooser(value);
     } else {
-        return windowPrompt(text);
+        return windowPrompt(value);
     }
 }
 
@@ -90,16 +89,16 @@ function simplemdeAttach(id) {
         autofocus: false,
     });
     mde.options.promptTexts = {
-        link: OVERRIDE_TEXT,
+        // We pass the editor instance as the prompt text for links in order to:
+        // - Override window.prompt in the appropriate scenarios.
+        // - Have access to the editor instance in our custom link chooser.
+        link: mde,
     };
     mde.render();
 
     mde.codemirror.on('change', function() {
         $('#' + id).val(mde.value());
     });
-
-    // TODO Make sure this is compatible with multiple instances of SimpleMDE on the page.
-    window.prompt = windowPromptOverride.bind(null, openLinkChooser.bind(null, mde));
 }
 
 // Refresh the markdown entry field when the tab buttons are clicked.
--- a/core/static/core/js/refresh_codemirror.js
+++ b/core/static/core/js/refresh_codemirror.js
@@ -1,5 +1,7 @@
-// Extracted from simplemde source:
-// https://github.com/sparksuite/simplemde-markdown-editor/blob/6abda7ab68cc20f4aca870eb243747951b90ab04/src/js/simplemde.js#L791-L822
+/**
+ * Extracted from simplemde source:
+ * https://github.com/sparksuite/simplemde-markdown-editor/blob/6abda7ab68cc20f4aca870eb243747951b90ab04/src/js/simplemde.js#L791-L822
+ */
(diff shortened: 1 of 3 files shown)


core/static/core/js/refresh_codemirror.js
```diff
@@ -1,22 +1,120 @@
...
+        var isPageLink = !!data.slug;
+        var url = isPageLink ? 'slug:' + data.slug : data.url;
+        _replaceSelection(cm, mde.getState().link, ['[', '](#url#)'], url);
+    }
+}
+
+function openLinkChooser(mde) {
+    var cm = mde.codemirror;
+
+    var workflow = window.ModalWorkflow({
```
thibaudcolas: This is relatively similar to the integration of Wagtail's default editor, except this integration doesn't support editing existing links. See https://github.com/wagtail/wagtail/blob/3134ffa119c15097c8b3c2a0422dd89f7c23304a/client/src/components/Draftail/sources/ModalWorkflowSource.js#L149-L166.

core/static/core/js/refresh_codemirror.js
```diff
@@ -1,22 +1,120 @@
...
+}
+
+function onLinkChosen(mde, data) {
+    var cm = mde.codemirror;
+
+    if (data.prefer_this_title_as_link_text) {
+        // Override the whole selection with link text + URL from the chooser.
+        cm.replaceSelection('[' + data.title + '](' + data.url + ')');
+    } else {
+        var isPageLink = !!data.slug;
```
thibaudcolas: The slug comes from the `data-` attribute in the chooser's template below.

core/widgets.py
```diff
@@ -20,6 +20,7 @@ def media(self):
             },
             js=(
                 'https://cdn.jsdelivr.net/simplemde/latest/simplemde.min.js',
-                'core/js/refresh_codemirror.js'
+                'core/js/refresh_codemirror.js',
+                'wagtailadmin/js/page-chooser-modal.js',
```
thibaudcolas: The page chooser modal's JS isn't loaded by default, it has to be included explicitly.

core/static/core/js/refresh_codemirror.js
```diff
@@ -1,22 +1,120 @@
...
+        onload: window.PAGE_CHOOSER_MODAL_ONLOAD_HANDLERS,
+        responses: {
+            pageChosen: function(data) {
+                onLinkChosen(mde, data);
+                workflow.close();
+            },
+        },
+        onError: function() {
+            window.alert('Uh-oh, the link chooser failed to open. Please refresh the page and try again.');
+        },
```
thibaudcolas: This is for AJAX errors only, so not much can be done except "retry".

ababic:
@thibaudcolas This is lovely! However, I am experiencing some pretty odd behaviour when attempting to update a link added by this chooser (the entire paragraph text seems to get duplicated).

ababic:
@thibaudcolas From what I can tell, this is an existing issue. It seems there is a wider problem with the markdown editor:

mde-dupe-content

@richtier What would you recommend we do here? Is it worth raising this as a separate bug/issue and working to resolve it outside this PR?

thibaudcolas:
👌 investigating this issue, it seems to be a bug in SimpleMDE's link insertion logic when there is a link to edit. It’s this particular branch of the code: https://github.com/sparksuite/simplemde-markdown-editor/blob/6abda7ab68cc20f4aca870eb243747951b90ab04/src/js/simplemde.js#L803-L811

I've reused this exact link insertion logic with the chooser UI, but there is nothing preventing us from rolling our own. @richtier would be good to hear your thoughts on whether this should happen as part of this PR, or separately.

richtier:
the key take home point for me is

>  From what I can tell, this is an existing issue. It seems there is a wider problem with the markdown editor

if it's an existing issue then I have no qualms with your PR being merged

it adds more value than not having it

thibaudcolas:
🤘 rebased and merged.

PR #361, opened 2019-03-28
Make the page reordering UI easier to find (TOR-95)

thibaudcolas:
Fixes [TOR-95](https://uktrade.atlassian.net/browse/TOR-95), coming from user research / feedback. The implementation overrides the corresponding parts of the Wagtail UI, re-implementing a PR to Wagtail itself: https://github.com/wagtail/wagtail/pull/5187.

The changes done here should be removed when upgrading to Wagtail v2.5, which includes this as well.

Here is what the new page reordering toggle looks like:

tor-95-reordering

Previous version for reference:

tor-95-before

There is also a hover/active state, using the same teal color as other columns in the header.

Tested in Chrome, Firefox, Safari on macOS. IE11 in Windows 8.1.


core/templates/wagtailadmin/pages/listing/_table_headers_explore.html
```diff
@@ -0,0 +1,67 @@
...
+                    <a href="{% url 'wagtailadmin_explore' parent_page.id %}" title="{% trans 'Disable ordering of child pages' %}">
+                        {% wagtail_icon name="order" %}{% trans 'Sort' %}
+                    </a>
+                {% else %}
+                    <a href="{% url 'wagtailadmin_explore' parent_page.id %}?ordering=ord" title="{% trans 'Enable ordering of child pages' %}">
+                        {% wagtail_icon name="order" %}{% trans 'Sort' %}
+                    </a>
+                {% endif %}
+            {% endif %}
+        </th>
```
thibaudcolas: This bit here is the only change compared to vanilla Wagtail 2.3. This might be easier to review by looking at the diff of the Wagtail PR, https://github.com/wagtail/wagtail/pull/5187/files#diff-d3c70845b0ed65756f583ff26d986644.

thibaudcolas:
Thanks @sdonk!

PR #378, opened 2019-04-05
Made capital invest banner title and image required and removed image captions as no longer used and added cta url fields for all cta

jessgilbert:
Made the most important fields required so it looks fine with the minimum amount of fields added 
image
And removed captions from the cms as captions would not be translated so removed them all together


invest/models.py
```diff
@@ -662,104 +660,41 @@ class InvestHomePage(ExclusivePageMixin, BasePage):
...
-                FieldPanel('subsection_content_two'),
+                FieldPanel('capital_invest_section_title'),
+                FieldPanel('capital_invest_section_content'),
+                FieldPanel('capital_invest_section_cta_text'),
+                FieldPanel('capital_invest_section_cta_url'),
+                ImageChooserPanel('capital_invest_section_background_image'),
             ],
-            heading='subsections two',
+            heading='Capital Investment section',
             classname='collapsible'
```
Kenneth-Siu: Putting these parameters at the top (and then using `child=[...]` for the `FieldPanels` will make these `MultiFieldPanel`s slightly more readable.

invest/migrations/0023_auto_20190409_0927.py
```diff
@@ -0,0 +1,349 @@
...
+            model_name='investhomepage',
+            name='setup_guide_call_to_action_url_ru',
+            field=models.URLField(max_length=255, null=True),
+        ),
+        migrations.AddField(
+            model_name='investhomepage',
+            name='setup_guide_call_to_action_url_zh_hans',
+            field=models.URLField(max_length=255, null=True),
+        ),
+    ]
```
Kenneth-Siu: For a single PR, try and keep to one migration. You can just delete this migrations and then run the `make migrations` again to squash them into a single migration.

Unless it gets too long for the CI system to cope with. Then you might have to split them up again. I think Alessandro ran into this before...

invest/tests/factories.py
```diff
@@ -37,18 +37,26 @@ class Meta:
...
+    capital_invest_section_background_image = factory.SubFactory(
+        wagtail_factories.ImageFactory
+    )
     sector_title = factory.fuzzy.FuzzyText(length=10)
     sector_button_text = factory.fuzzy.FuzzyText(length=10)
+    sector_button_url = 'http://foo.com'
     hpo_title = factory.fuzzy.FuzzyText(length=10)
     setup_guide_title = factory.fuzzy.FuzzyText(length=10)
     setup_guide_call_to_action_text = factory.fuzzy.FuzzyText(length=10)
+    setup_guide_call_to_action_url = 'http://foo.com'
```
Kenneth-Siu: I think you can just use the same fuzzy test generator as for the other text fields?

jessgilbert: This will be changed back to fuzzy text now as they aren't url fields

PR #384, opened 2019-04-08
Revert "Fix url generation for legacy pages"

sdonk:


richtier:
will try and fix instead of reverting

PR #389, opened 2019-04-09
Fix HPO urls

richtier:


richtier:
this is the hotfix that went into release already

PR #413, opened 2019-04-19
[CMS-13998] Replace lookup_by_full_path with lookup_by_path

ababic:
Adds `PageLookupByPathAPIEndpoint` view, url and related tests

Removes `PageLookupByFullPathAPIEndpoint` view, url and related tests

Changes `PageLookupBySlugAPIEndpoint` to use the same base class as `PageLookupByPathAPIEndpoint` and only override what it needs to.

Adds `PageIDCache` to serve both `PageLookupByPathAPIEndpoint` and `PageLookupBySlugAPIEndpoint` endpoints.

Changes to `PageCache`:
- Keys just use the page id and any additional variation values (e.g. language/region).
- The order of variation key/values passed to the key builder method no longer matters.

Changes to `CachePopulator` and `RegionAwareCachePopulator`:
- Added a dedicated `delete()` method for each class
- Updated cache set/delete operations to be compatible with the new `PageCache` key builder method

Changes to `AbstractDatabaseCacheSubscriber`:
- Change `delete()` to utilise the `delete()` method from the relevant `cache_populator`, instead of implementing it's own 'non region aware' version


core/tests/test_views.py
```diff
@@ -314,92 +440,25 @@ def test_lookup_by_slug_missing_required_query_param(translated_page,
...
+    slug = 'thing'
 
-    response = admin_client.get(url, {'service_name': cms.FIND_A_SUPPLIER})
-
-    assert response.status_code == 404
+    url = reverse('api:lookup-by-slug', kwargs={'slug': slug})
 
-
-@pytest.mark.django_db
-def test_lookup_by_slug_draft(page_with_reversion, client):
```
ababic: The updated `test_api_serves_drafts()` test nearer the top of this module has this endpoint covered, so deleting this one

core/tests/test_views.py
```diff
@@ -314,92 +440,25 @@ def test_lookup_by_slug_missing_required_query_param(translated_page,
...
-    )
-
-    assert published_response.status_code == 200
-    assert published_data['title'] == 'published-title'
-    assert published_data['meta']['url'] == page_with_reversion.get_url()
-
-
-@pytest.mark.django_db
-@patch('core.filters.ServiceNameFilter.filter_service_name')
-def test_lookup_by_slug_filter_called(mock_filter_service_name, admin_client):
```
ababic: Thanks to the new `PageIDCache`, the view no longer needs to use `core.filters.ServiceNameFilter.filter_service_name`, so this test is no longer relevant

core/tests/test_views.py
```diff
@@ -59,51 +61,129 @@ def test_permissions_published(rf):
...
+    query_params = {'service_name': 'FIND_A_SUPPLIER'}
+    query_params.update(languge_query_params)
+    response = client.get(url, query_params)
     assert response.status_code == 200
     assert response.json()['title'] == 'ENGLISH'
 
 
 @pytest.mark.django_db
-def test_api_draft(client, page_with_reversion):
+def test_api_serves_drafts(
```
ababic: This test now covers all main endpoints to ensure they function the same way

core/tests/test_views.py
```diff
@@ -59,51 +61,129 @@ def test_permissions_published(rf):
...
+    assert response.json()['title'] == expected_title
 
 
 @pytest.mark.parametrize('language_code', (
     'en-gb' 'de' 'ja', 'zh-hans', 'fr', 'es', 'pt', 'ar',
 ))
 @pytest.mark.django_db
-def test_api_translations_not_populated(
-    client, untranslated_page, language_code
+def test_api_falls_back_to_english_when_translations_unavailable(
```
ababic: This test now covers all main endpoints to ensure they function the same way

core/tests/test_views.py
```diff
@@ -59,51 +61,129 @@ def test_permissions_published(rf):
     ('ar', 'ARABIC'),
 ))
 @pytest.mark.django_db
-def test_api_translations(client, translated_page, language_code, expected):
+def test_api_translations_are_loaded_when_available(
```
ababic: This test now covers all main endpoints to ensure they function the same way

core/cache.py
```diff
@@ -22,28 +21,39 @@ class PageCache:
...
+        variation_kwargs_sorted = sorted(
+            item for item in variation_kwargs.items()
+            if item[1]
+        )
+
+        # improve reliability of delete_many() by creating a redis hashtag
+        # from `page_id`. This ensures keys related to the same page are stored
+        # in the same node in a clustered environment
+        hashtag_val = f'page-{page_id}'
+        return '{{%s}}%s' % (hashtag_val, urlencode(variation_kwargs_sorted))
```
richtier: good news is this can be simplified because we're using unclustered redis, so the hashtag bit cna be removed.

clustered redis sucks for the hoops we have to jump through for batch operations, and celery does not like - so we moved to unclustered.

core/cache.py
```diff
@@ -22,28 +21,39 @@ class PageCache:
...
+        """
+        Return a string from the supplied arguments that is suitable for use
+        as a cache key.
+
+        Vales are prepended with page-specific redis hashtag to improve
+        reliability of delete_many() in clustered cache configurations.
+        """
+        variation_kwargs_sorted = sorted(
+            item for item in variation_kwargs.items()
+            if item[1]
```
richtier: `canonicalize_url` does this and a few other things: https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.canonicalize_url

why not use `canonicalize_url`?

core/cache.py
richtier: oooh. now that the prefix is needed - `page_id` could be included in `variation_kwargs` :)

core/cache.py
```diff
@@ -120,14 +122,20 @@ def populate_sync(cls, instance):
...
-                            'lang': language_code,
-                            'service_name': instance.service_name,
-                        },
+                        page_id=instance.id,
                         data=serializer.data,
+                        lang=language_code,
                     )
 
+    @classmethod
+    def delete(cls, instance):
```
richtier: it's a bit off for a class called `CachePopulator` to have a method called `delete`.

feel free to rename `CachePopulator` to something more fitting. Naming is hard :laughing:

core/cache.py
```diff
@@ -228,11 +218,140 @@ def populate_sync(cls, instance):
...
+        key = cls.build_slug_lookup_key(service_name, slug)
+        return cls.get_mapping(cls.by_slug_map_key).get(key)
+
+    @classmethod
+    def get_for_path(cls, path, site_id):
+        key = cls.build_path_lookup_key(site_id, path)
+        return cls.get_mapping(cls.by_path_map_key).get(key)
+
+    @classmethod
+    def subscribe_to_signals(cls):
```
richtier: for consistency with existing could this be `subscribe`? Either that or rename existing `subscribe_to_signals` :)

core/cache.py
```diff
@@ -228,11 +218,140 @@ def populate_sync(cls, instance):
...
+ROOT_PATHS_TO_SERVICE_NAMES = {
+    'export-readiness-app': cms.EXPORT_READINESS,
+    'great-international-app': cms.GREAT_INTERNATIONAL,
+    'find-a-supplier-app': cms.FIND_A_SUPPLIER,
+    'invest-app': cms.INVEST,
+    'components-app': cms.COMPONENTS,
+}
+
+
+class PageIDCache:
```
richtier: why is this a property of `AbstractDatabaseCacheSubscriber`? could it be a first class citizen at the module level?

core/cache.py
ababic: > we're using unclustered redis now, so the hashtag prefix can be removed.

I'm all for removing code. But, although this was added specifically for clustered redis, I wonder if it might be best left in-place, in case the architecture choice changes again? I think this same approach might also work for clustered Memcache or other backends. While we're still using `delete_many()` here (which won't be much longer), I'd propose that we leave it in place for now.

> now that the prefix is no onger needed - page_id could be included in variation_kwargs :)

The **kwargs device is cool for variable numbers of unknown optional arguments, but since `page_id` is required, I think the current design makes sense from a design/readability perspective.

core/cache.py
ababic: I must say, I'm split in this one. If all we ever had was a url string with the arguments already combined, and we were actually using a url format as a key, then `canonicalize_url` would be more obviously useful. However, since `variation_kwargs` is always a dictionary, achieving what we want with a single line of expressive, readable python seems like a better choice, and likely more efficient to boot.

core/cache.py
ababic: I would say this naming is just about as sensible as it could be :) The class's primary purpose is population, but clearing has always been a part of that. The only difference now is that the logic is in a separate method.  Could we I suggest that we leave things as they are for now? We'll undoubtedly be revisiting this area again soon, anyhow.

core/cache.py
ababic: Is this a reference to `ROOT_PATHS_TO_SERVICE_NAMES`? If so, I beleive it already is :) happy to move it to the top of the module though, as is usually the done thing for globals.

core/cache.py
ababic: I'm a descriptive naming freak, so naturally prefer `subscribe_to_signals`. However, for the sake of keeping this PRs change footprint down, I may just go with `subscribe` here.

core/cache.py
ababic: Resolved in 09d15b5

core/cache.py
ababic: Resolved in 0e1532d

core/cache.py
ababic: @richtier Also, I don't know as we need any of the other stuff that `canonicalize_url` does. We're only really urlencoding the values here because it is an inexpensive way to ensure keys don't contain spaces or other special characters that Memcached and other backends will choke on.

core/cache.py
ababic: Hashtagging removed in 1fcb7fd

ababic:
@sdonk CHANGELOG updated in 6c5427e

PR #419, opened 2019-04-24
Removed now unused fields for capital investment section

jessgilbert:
The component used for this section has changed and now the image and cta are not used, it is only the title and some text


jessgilbert:
Sorry the branch name should start with CI not CL to point to the correct jira ticket

PR #424, opened 2019-04-26
Added Investment support directory section to invest home page

jessgilbert:


invest/serializers.py
```diff
@@ -209,6 +209,10 @@ class InvestHomePageSerializer(BasePageSerializer):
         max_length=255,
         allow_null=True
     )
+    isd_section_image = wagtail_fields.ImageRenditionField('original')
```
Kenneth-Siu: Might be worth chatting to Edie about this? IIRC, she has some thoughts about how we store and serve images in the CMS.

PR #427, opened 2019-04-30
[CMS-1397] Optimise get_tree_based_url() by fetching routing settings and site simultaneously

ababic:
Technically, this is covered by the existing `CMS-1397` entry in the CHANGELOG, but please let me know if you'd like me to add something else


ababic:
@sdonk CHANGELOG updated in 079d81a

ababic:
@sdonk I honestly don't know what this CHANGLOG is supposed to look like now. If we're going to add changelog entries as part of each PR, they need to be added literally just before merging. If not, this will happen all the time :(

PR #431, opened 2019-05-01
[CMS-1446] Prevent changes to user details

kaedroho:
This disables the following:

 - All basic user fields in the users admin interface (and completely removes password fields)
 - Ability for users to change their own password or email address
 - Password reset


users/forms.py
```diff
@@ -44,10 +44,46 @@ def get_groups_queryset(self, user):
...
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        instance = getattr(self, 'instance', None)
+        if instance and instance.pk:
+            # Disable the fields
+            self.fields['username'].widget.attrs['readonly'] = True
+            self.fields['email'].widget.attrs['readonly'] = True
+            self.fields['first_name'].widget.attrs['readonly'] = True
```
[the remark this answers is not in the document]
kaedroho: It's quite a bit tidier that way, thanks for spotting!

users/forms.py
```diff
@@ -44,10 +44,27 @@ def get_groups_queryset(self, user):
...
-class UserEditForm(EntryPointAwareUserActionForm, wagtail_forms.UserEditForm):
+class DisableUserFieldsMixin(forms.Form):
+    """
+    Disables all user fields except for roles
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        instance = getattr(self, 'instance', None)
+        if instance and instance.pk:
```
sdonk: Can an instance exist and not have a pk?

conf/settings.py
```diff
@@ -462,3 +462,9 @@
 ACTIVITY_STREAM_ACCESS_KEY_ID = env.str('ACTIVITY_STREAM_ACCESS_KEY_ID')
 ACTIVITY_STREAM_SECRET_ACCESS_KEY = \
     env.str('ACTIVITY_STREAM_SECRET_ACCESS_KEY')
+
+
+# Disable password management in Wagtail admin
+WAGTAIL_PASSWORD_MANAGEMENT_ENABLED = False
+WAGTAIL_PASSWORD_RESET_ENABLED = False
+WAGTAILUSERS_PASSWORD_ENABLED = False
```
sdonk: Although they're unlike to change, these should probably come from env vars which default to False. Just in case we need to enable/disable it for any reason

users/forms.py
kaedroho: No, ``pk`` is always an alias to the primary key field which must be set in order for an instance to exist.

conf/settings.py
kaedroho: I doubt we will ever re-enable password management on the servers because that would mean we'll have to disable SSO. But it might be nice to have it enabled for local development, so will make this configurable! But I think having it enabled by default would make more sense as SSO is usually disabled by default.

kaedroho: Or, I could make it check whether SSO is configured and disable Wagtail password management if so?

users/forms.py
sdonk: Why the double check then?

conf/settings.py
sdonk: SSO is currently deployed but disabled on dev. It's disabled by default on local and enabled by default on the envs. There's an env var called enforce_staff_sso

users/forms.py
ababic: @sdonk When `UserCreationForm` runs this code, `instance` will be `None` - so not checking `instance` first would result in an `AttributeError`

conf/settings.py
```diff
@@ -478,3 +478,12 @@
 ACTIVITY_STREAM_ACCESS_KEY_ID = env.str('ACTIVITY_STREAM_ACCESS_KEY_ID')
 ACTIVITY_STREAM_SECRET_ACCESS_KEY = \
     env.str('ACTIVITY_STREAM_SECRET_ACCESS_KEY')
+
+
+# Disable password management in Wagtail admin
+WAGTAIL_PASSWORD_MANAGEMENT_ENABLED = env.str(
+    'WAGTAIL_PASSWORD_MANAGEMENT_ENABLED', False)
```
sdonk: Should they go alongside with `ENFORCE_STAFF_SSO_ON`?

PR #438, opened 2019-05-07
[NOTICKET] Performance improvements

kaedroho:
**Rewrite some page specific queries to be faster**

Instead of filtering a generic Page queryset by page location and then fetching specific versions of the pages, query on the specific model directly and filter them by page location. This way, Wagtail only needs to perform a single query.

**Improve performance of BasePage.full_path**

This removes the specific page lookups from this method, which are not necessary. This method is called a lot, so this improves performance very noticibly when rendering pages that have a lot of page links.


ababic:
@sdonk Changelog updated in ec2266f :)

PR #439, opened 2019-05-07
CMS-1417 Integrate staff SSO

sdonk:


conf/urls.py
```diff
@@ -118,6 +118,19 @@
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 
+if settings.FEATURE_FLAGS['ENFORCE_STAFF_SSO_ON']:
+    urlpatterns = [
+        url('^sso/logged-in-landing/$',
+            SSOLoggedinLandingView.as_view(), name='sso_logged_in_landing'),
```
richtier: is it feasible to use a feature flag check in `dispatch` to make testing this feature flag easier?

sdonk: The SSO URLs are added here based on the feature flag.
I'd prefer to have one point only where the feature flag does things, instead of being checked in multiple places.

PR #441, opened 2019-05-08
CMS-1449 Upgrade to wagtailmedia 0.3.0, for Wagtail 2.4+ compatibility

thibaudcolas:
See [CMS-1449](https://uktrade.atlassian.net/browse/CMS-1449).

This should make it easier to upgrade to a more recent Wagtail version once we get there. The 0.3.0 release does not introduce any changes except for Wagtail 2.4 & 2.5 compatibility, but APIs used by directory-cms haven't changed. The released changes have been tested against this project with Django 1.11 and Wagtail 2.3, and also a regular Wagtail build with Django 2.2 and Wagtail 2.6-dev (latest changes from Wagtail master).

I've tested this manually by interacting with the Media section of the admin, and the media chooser modal. Everything seems to be working as expected. Also ran through a longer list of manual tests before making the release, https://github.com/torchbox/wagtailmedia/pull/43.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [CMS-1417](https://uktrade.atlassian.net/browse/CMS-1417) Integrate staff SSO
 - [CMS-1462](https://uktrade.atlassian.net/browse/CMS-1462) Remove char limit from fields in `InternationalArticlePage`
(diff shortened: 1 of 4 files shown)


thibaudcolas:
👌 my bad, should be done now.

PR #450, opened 2019-05-10
[NOTICKET] Remove blanket logging

kaedroho:
This removes some noise when using the shell. The current set up means a lot of unneeded logs are being shown in the shell at inconvenient times making interactive UIs such as ``ipython`` and ``pdb`` difficult to use.


kaedroho:
I think we can ignore that. This doesn't add or change any code

PR #451, opened 2019-05-10
[CMS-1421] Integrate with GOV.UK Notify for SSO user approval notifications

ababic:
The functions in `core.notfications` here will eventually be called by `SSORequestAccessView. notify_team_leader_of_pending_approval()` and `EditUserView.notify_user_of_approval()` in PR #443.

If you'd like to test them from the shell, you'll need to copy the necessary env vars from the key vault from dev.... I'll also need to add your email to a whitelist so that you can receive the email from the Notify service.


makefile
```diff
@@ -90,7 +90,10 @@ TEST_SET_ENV_VARS := \
 	export EMAIL_HOST_USER=debug; \
 	export EMAIL_HOST_PASSWORD=debug; \
 	export DEFAULT_FROM_EMAIL=debug; \
-	export FEATURE_ENFORCE_STAFF_SSO_ENABLED=false
+	export FEATURE_ENFORCE_STAFF_SSO_ENABLED=false; \
+	export GOVNOTIFY_API_KEY=directory_cms_test-03185ee5-578c-4ffc-8774-2288e7b34e63-e82262ea-ae8c-4c6d-8570-c16afdc8347f; \
```
ababic: FYI: This is NOT a valid key... we just need a key in a vaild format, otherwise the notifications client will not initialise here: https://github.com/uktrade/directory-cms/pull/451/files#diff-258a758f52302ff9ddd2aa30614b3172R5

sdonk: can you stick it in a note somewhere?

ababic: @sdonk I have resolved this in 0a55d83 by adding a correctly-formatted default in `conf.settings` instead, and added a comment there.

ababic:
@sdonk. I'm just going to adjust this PR to connect the necessary views to these :)

PR #454, opened 2019-05-21
Parse authbroker url as string

sdonk:


conf/settings.py
```diff
@@ -450,7 +450,7 @@
     LOGIN_REDIRECT_URL = reverse_lazy('wagtail_homepage')
 
     # authbroker config
-    AUTHBROKER_URL = env.url('STAFF_SSO_AUTHBROKER_URL')
+    AUTHBROKER_URL = env.str('STAFF_SSO_AUTHBROKER_URL')
```
sdonk: urljoin doesn't like this to be a ParserURLResult

PR #457, opened 2019-05-22
Ci 104 capital invest

jessgilbert:


great_international/apps.py
```diff
@@ -17,6 +17,10 @@ def ready(self):
         cache.InternationalGuideLandingPageSubscriber.subscribe()
         cache.InternationalEUExitFormPageSubscriber.subscribe()
         cache.InternationalEUExitFormSuccessPageSubscriber.subscribe()
+        cache.InternationalCapitalInvestLandingPageSubscriber.subscribe()
+        cache.CapitalInvestRegionPageSubscriber.subscribe()
+        cache.CapitalInvestRegionalSectorPageSubscriber.subscribe()
+        cache.CapitalInvestOpportunityPageSubscriber.subscribe()
```
richtier: :sunglasses:

great_international/cache.py
```diff
@@ -75,3 +75,43 @@ class InternationalEUExitFormSuccessPageSubscriber(
 ):
     model = models.InternationalEUExitFormSuccessPage
     subscriptions = []
+
+
+class InternationalCapitalInvestLandingPageSubscriber(
```
richtier:
```
class InternationalCapitalInvestLandingPageSubscriber(
    AbstractDatabaseCacheSubscriber
):
```

would be slightly easier to read :)

core/serializers.py
```diff
@@ -54,3 +54,14 @@ def get_child_pages_data_for(self, parent, page_type, serializer):
             context=self.context
         )
         return serializer.data
+
+
+class ParentPageSerializerHelper(serializers.Serializer):
+    def get_parent_page(self, instance, serializer):
```
MichaelWalker: tiny point... 😊

The ChildPageSerializer names its method `get_child_pages_data_for`.
I wonder if we should follow that convention and have `get_parent_page_data_for`?

great_international/cache.py
```diff
@@ -75,3 +75,44 @@ class InternationalEUExitFormSuccessPageSubscriber(
...
+):
+    model = models.InternationalCapitalInvestLandingPage
+    subscriptions = [models.CapitalInvestRegionPage]
+
+
+class CapitalInvestRegionPageSubscriber(AbstractDatabaseCacheSubscriber):
+    model = models.CapitalInvestRegionPage
+    subscriptions = [
+        models.CapitalInvestRegionalSectorPage,
+        models.InternationalCapitalInvestLandingPage
```
MichaelWalker: What do we need this subscription for?

MichaelWalker: (Similarly with the other subscribers below)

I'd expect to only have to subscribe to pages whose content affects the content on this page, which is normally just children (or related pages) that we want to link to.

But we often seem to also include a subscription to the parent pages. Is this used for anything?

core/serializers.py
jessgilbert: Good point, will change that :)

great_international/cache.py
jessgilbert: Yep, don't need that will delete

great_international/models.py
```diff
@@ -1709,3 +1716,1165 @@ class InternationalEUExitFormSuccessPage(ExclusivePageMixin, BasePage):
...
+        verbose_name="Region opportunities section title",
+        blank=True
+    )
+    region_ops_section_intro = models.CharField(
+        max_length=255,
+        blank=True,
+        verbose_name="Region opportunities section intro"
+    )
+
+    related_region_one = models.ForeignKey(
```
MichaelWalker: Perhaps not worth doing right now, but would it be slightly nicer to use the Orderable pattern again here so that editors can add as many regions as they like?

great_international/models.py
```diff
@@ -1709,3 +1716,1165 @@ class InternationalEUExitFormSuccessPage(ExclusivePageMixin, BasePage):
...
+    ]
+
+    edit_handler = make_translated_interface(
+        content_panels=content_panels,
+        settings_panels=settings_panels
+    )
+
+
+class AddedRelatedPageSummary(models.Model):
+    added_related_pages = models.ForeignKey(
```
MichaelWalker: I think the naming here might be slightly misleading.

AddedRelatedPageSummary isn't really a summary - its a link to the entire page, and it isn't really as general as it claims (as it can only really be a OpportunityPage).

Also `added_related_pages` isn't actually a list, its a reference to a single opportunity page. 

perhaps something like the below would make it a little clearer what is going on?
```
class RelatedOpportunity(models.Model):
    opportunity = models.ForeignKey(stuff)
    panels = otherStuff

class SectorRelatedOpportunities(Orderable, RelatedOpportunity):
    page = ParentalKey(stuff)

class CapitalInvestRegionalSectorPage(BasePage):
    # yet more stuff
```

great_international/serializers.py
```diff
@@ -167,13 +287,74 @@ class RelatedCampaignPageSerializer(BasePageSerializer):
...
-    InternationalCampaignPage: RelatedCampaignPageSerializer,
-}
+        InternationalArticlePage: RelatedArticlePageSerializer,
+        InternationalCampaignPage: RelatedCampaignPageSerializer,
+        CapitalInvestOpportunityPage:
+        RelatedCapitalInvestOpportunityPageSerializer,
+    }
+
+
+class PageWithRelatedRegionPagesSerializer(serializers.Serializer):
```
MichaelWalker: We only seem to use this class once at the moment. Are we expecting to use it again lots?

If not I'd be tempted to just move this method into `InternationalCapitalInvestLandingPageSerializer` and get rid of this class.

great_international/serializers.py
```diff
@@ -167,13 +287,74 @@ class RelatedCampaignPageSerializer(BasePageSerializer):
...
+        serialized = []
+        items = [
+            obj.related_region_one,
+            obj.related_region_two,
+            obj.related_region_three,
+            obj.related_region_four,
+            obj.related_region_five,
+            obj.related_region_six,
+        ]
+        for related_region in items:
```
MichaelWalker: If you've not seen python's [list comprehension](https://www.pythonforbeginners.com/basics/list-comprehensions-in-python) before, then its really cool!

for example, this would become something like
```
return [RelatedCapitalInvestPageSerializer(region.specific) for region in items if region is not None]
```

MichaelWalker: (this makes a rather long line in this case, so up to you what you go with, but worth having a play around and knowing that it exists! 😊

great_international/serializers.py
```diff
@@ -647,3 +826,337 @@ class InternationalEUExitFormSuccessPageSerializer(BasePageSerializer):
...
+
+    contact_section_title = serializers.CharField(max_length=255)
+    contact_section_text = serializers.CharField(max_length=255)
+    contact_section_cta_text = serializers.CharField(max_length=255)
+
+    region_cards = serializers.SerializerMethodField()
+
+    def get_region_cards(self, instance):
+        data = [
+            RegionCardsProxyDataWrapper(
```
MichaelWalker: Clever 😊

PR #462, opened 2019-05-24
Ci 144 update capital invest landing page

jessgilbert:


great_international/serializers.py
```diff
@@ -308,6 +258,37 @@ class RelatedCapitalInvestOpportunityPageSerializer(BasePageSerializer):
...
+    def get_related_region(self, obj):
+        serialized = []
+        region = obj.related_region
+
+        if not region:
+            return serialized
+        serializer = RelatedCapitalInvestPageSerializer(
+            region.specific)
+        serialized.append(serializer.data)
+        return serialized[0]
```
sdonk: Can you not just return `serializer.data`?

jessgilbert: I can indeed you're completely right, will change them :)

great_international/models.py
```diff
@@ -1718,207 +1718,183 @@ class InternationalEUExitFormSuccessPage(ExclusivePageMixin, BasePage):
...
-        max_length=255,
-        blank=True
-    )
-    reason_to_invest_section_intro = models.CharField(
-        max_length=255,
-        blank=True
-    )
-    reason_to_invest_section_content = MarkdownField(blank=True)
-    reason_to_invest_section_image = models.ForeignKey(
+class RegionCardField(models.Model):
```
sdonk: Any particular reason why this is not inheriting from `BasePage`

great_international/models.py
```diff
@@ -1718,207 +1718,183 @@ class InternationalEUExitFormSuccessPage(ExclusivePageMixin, BasePage):
...
+        related_name='added_region_card_fields',
         blank=True,
-        on_delete=models.SET_NULL,
-        related_name='+'
+        null=True,
     )
-    region_card_two_image = models.ForeignKey(
+
+
+class HomesInEnglandCardField(models.Model):
```
sdonk: ditto

PR #464, opened 2019-05-29
Ci 150 change to flexible cards in cms

jessgilbert:
Two migrations required as removing fields that were previously non-nullable so had to make them blank and migrate before removing them and migrating. This is so it can be rolled back.


invest/serializers.py
```diff
@@ -197,17 +237,20 @@ class InvestHomePageSerializer(BasePageSerializer):
...
+    def get_featured_cards(self, instance):
+        data = [
+            FeaturedCardsDataWrapper(
+                instance=instance,
+                position_number=num
+            )
+            for num in ['one', 'two', 'three']
+        ]
+        serializer = FeaturedCardsSerializer(data, many=True)
+        return serializer.data
```
richtier: could you please move the method below the fields - the dev reading this would probably get to `get_featured_cards` and (incorrectly) conclude the fields have ended

invest/migrations/0029_auto_20190529_1039.py
```diff
@@ -0,0 +1,694 @@
...
+            model_name='investhomepage',
+            name='featured_card_two_title_pt',
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.AddField(
+            model_name='investhomepage',
+            name='featured_card_two_title_zh_hans',
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.AlterField(
```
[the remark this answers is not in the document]
jessgilbert: It is deleted in the next migration, I had to do it in two as it was a non-nullable field so had to change it to be able to be blank and migrate and then delete and migrate. Otherwise I wouldn't be able to roll back which is unhelpful when working on different branches on the cms

invest/serializers.py
jessgilbert: Done

PR #473, opened 2019-05-31
CMS-1591 move invest models in international app

sdonk:


ababic:
Hi @eadpearce. There is now a `BaseInternationalPage` in `great_international.models` that these new models should subclass instead of `BasePage`.  You should also be able to remove the `service_name_value = cms.GREAT_INTERNATIONAL` line from each of them too.

PR #474, opened 2019-06-04
[CMS-1603] Remove unique slug enforcement for international

ababic:
This is one in a series of changes that will help us to migrate from modeltranslation to page-level translation. 

- Moved unique slug inforcement logic from `BasePage` to a separate `ServiceNameUniqueSlugMixin` class (dropping the `delete()` override completely, as it seems to be unnecessary, and is causing other issues (see CMS-1633)
- `BasePage` does not use `ServiceNameUniqueSlugMixin` by default
- Each individual app defines its own base page class (most using `ServiceNameUniqueSlugMixin`), which also sets the `service_name_value` attribute for that app (more DRY)
- `great_international.BaseInternationalPage` does not use `ServiceNameUniqueSlugMixin`.
- Refactored `core.filters.ServiceNameFilter` to work in a more efficient way, that doesn't depend on the model being a direct subclass of `BasePage`


core/filters.py
```diff
@@ -17,17 +11,23 @@ class Meta:
...
+            ctype.id for ctype in
+            ContentType.objects.get_for_models(*relevant_models).values()
         ]
-        queries = (
-            Q(**{concrete_model_name + '__service_name': value})
-            for concrete_model_name in concrete_model_names
-        )
-        return queryset.filter(reduce(operator.or_, queries))
+        # Filter the queryset
+        return queryset.filter(content_type_id__in=content_type_ids)
```
richtier: I'm glad you could simplify this optimization/mess I introduced :smile:

PR #486, opened 2019-06-10
Ci 165 redesign of capital invest pages

jessgilbert:


core/serializers.py
```diff
@@ -65,3 +65,32 @@ def get_parent_page_data_for(self, instance, serializer):
             context=self.context
         )
         return serializer.data
+
+
+class PageRelatedPageOfSelfSerializerHelper(serializers.Serializer):
+    def get_page_with_related_page_of_self_data_for(
```
MichaelWalker: The name makes it sound like this is quite a general class (ie would work for any type of page), but it looks like it would only actually work for finding opportunity pages related to a sector.

Could we 
1) change the name to make it reflect this
2) after that we won't need the page_type param either.

great_international/models.py
```diff
@@ -396,9 +404,14 @@ class Meta:
             heading='Case Study',
             classname='collapsible',
             children=[
+                HelpPanel('Required fields for section to show: '
+                          'Case Study Image, Case Study Title'),
+                ImageChooserPanel('case_study_image'),
                 FieldPanel('case_study_title'),
                 FieldPanel('case_study_description'),
                 FieldPanel('case_study_cta_text'),
+                HelpPanel('Cta\'s require both text and a link to show '
```
MichaelWalker: the `\'` works fine, but FYI, you can also use the other quotation marks, ie `"CTA's"` or `'"Hello World!"'` and again everything will work. (I've seen this approach more commonly than the escaping approach you went with, but they should both work fine)

great_international/models.py
```diff
@@ -2097,18 +2129,13 @@ class InternationalCapitalInvestLandingPage(
...
 
     breadcrumbs_label = models.CharField(max_length=255)
     hero_title = models.CharField(max_length=255)
     hero_image = models.ForeignKey(
         'wagtailimages.Image',
-        null=True,
-        on_delete=models.SET_NULL,
         related_name='+',
-        blank=True
+        on_delete=models.PROTECT
```
MichaelWalker: Why do we need this change?

jessgilbert: I changed it to not be allowed to be null and so the `on_delete=set.NULL` wouldn't work and it told me to set it at this instead

core/serializers.py
```diff
@@ -65,3 +65,32 @@ def get_parent_page_data_for(self, instance, serializer):
...
+
+        if not queryset:
+            return []
+
+        serializer = serializer(
+            queryset,
+            allow_null=True,
+            context=self.context
+        )
+        return serializer.data
```
MichaelWalker: would require some renaming... but, if we change this to `return serializer.data['opportunities']` then we can get rid of the slightly awkward `sector['related_opportunities']['opportunities']` thing that we currently have to do in the front end, and the data model becomes a little more natural and easy to work with.

great_international/serializers.py
```diff
@@ -339,20 +337,34 @@ def get_related_region(self, obj):
...
 
+class RelatedOpportunitySerializer(serializers.Serializer):
+    opportunities = serializers.SerializerMethodField()
+
+    def get_opportunities(self, obj):
+
+        serialized = []
+        for opp in obj:
+            if not opp:
+                return []
```
MichaelWalker: Do we want `continue` rather than `return []` here?

great_international/serializers.py
```diff
@@ -339,20 +337,34 @@ def get_related_region(self, obj):
...
-
+        serializer = MinimalPageSerializer(
+            sector.specific)
         return serializer.data
 
 
+class RelatedOpportunitySerializer(serializers.Serializer):
+    opportunities = serializers.SerializerMethodField()
+
+    def get_opportunities(self, obj):
```
MichaelWalker: Could we rename `obj`. Its not very clear to me what it represents.

great_international/serializers.py
```diff
@@ -497,6 +513,18 @@ def get_section_three_subsections(self, instance):
...
+
+    related_opportunities = serializers.SerializerMethodField()
+
+    def get_related_opportunities(self, obj):
+        opportunities = self.get_page_with_related_page_of_self_data_for(
+            obj,
+            CapitalInvestOpportunityPage,
+            RelatedOpportunitySerializer
+        )
+        return opportunities
```
MichaelWalker: just 
```python
return self.get_page_with_related_page_of_self_data_for(
            obj,
            CapitalInvestOpportunityPage,
            RelatedOpportunitySerializer
        )
``` ?

File: great_international/serializers.py [hunk 8]
```diff
@@ -339,20 +337,34 @@ def get_related_region(self, obj):
...
+    def get_opportunities(self, obj):
+
+        serialized = []
+        for opp in obj:
+            if not opp:
+                return []
+            serializer = RelatedCapitalInvestOpportunityPageSerializer(
+                opp.specific)
+            serialized.append(serializer.data)
+        return serialized
```
MichaelWalker: In other places we do something like
```python
serializer = <Foo>Serializer(
    <list of foo>,
    many=True,
    # other stuff
)
```

Does that not work in this case?

jessgilbert: Yes it does, have changed it :) 👍

core/serializers.py
```diff
@@ -65,3 +66,34 @@ def get_parent_page_data_for(self, instance, serializer):
             context=self.context
         )
         return serializer.data
+
+
+class OpportunityPageWithRelatedSectorPageSerializerHelper(
```
MichaelWalker: We only use this once, (and I don't think we are likely to reuse it) so I'd be tempted to just kill this helper and put the method directly into the serializer.

PR #487, opened 2019-06-10
Serialize breadcrumbs for GA360 page categories

eadpearce:


export_readiness/tests/test_serializers.py
```diff
@@ -87,3 +87,30 @@ def test_country_guide_page_serializer(root_page, rf):
     )
 
     assert len(serializer.data['intro_ctas']) == 0
+
+
+@pytest.mark.django_db
+def test_breadcrumbs_serializer(root_page, rf, django_assert_num_queries):
```
MichaelWalker: Really nice to have this in the CMS! :)

core/models.py
```diff
@@ -218,6 +218,15 @@ def get_url(self, is_draft=False, language_code=settings.LANGUAGE_CODE):
             url += '?' + urlencode(querystring)
         return url
 
+    @property
+    def ancestor_pages(self):
```
MichaelWalker: nice to have a property for this, but I think I'd be surprised that it did something different to `get_ancestors()` (ie, I don't think I'd expect it to exclude the root page etc from the name alone, as those are also pages)

Is there another name that we could think of that makes it a little more obvious what is going to happen?

core/serializers.py
```diff
@@ -14,6 +19,12 @@ class BasePageSerializer(serializers.Serializer):
     last_published_at = serializers.DateTimeField()
     title = serializers.CharField()
     page_type = serializers.SerializerMethodField()
+    tree_based_breadcrumbs = serializers.SerializerMethodField()
+
+    def get_tree_based_breadcrumbs(self, instance):
+        breadcrumbs = [page.specific for page in instance.ancestor_pages]
+        breadcrumbs.append(instance)
+        return PageTitleAndUrlSerializer(breadcrumbs, many=True).data
```
MichaelWalker: Am I right in thinking that this won't find the 'Home' page?

ie, if we have a site at `Great.gov.uk International -> industries -> energy`
then this will find the `industries -> energy` bit, but won't know anything about the `Great.gov.uk International` part?

Is there a nice way to also include the home page in this?
(on a related note, why isn't the home page of an app the parent of the other pages?)

core/models.py
eadpearce: yeah i'm not sure. unless we call it `ancestors_excluding_app_and_root` but i think that's a little too long >> maybe `page_tree`? but then that's not plural which may be confusing. `tree_pages`?

core/serializers.py
eadpearce: yeah i was thinking about this as well. eventually i think we want to convert the app pages into homepages so then we'll be able to do this but until then i'm leaving it out because i'm not sure

core/serializers.py
```diff
@@ -3,6 +3,11 @@
 from core import fields
 
 
+class PageTitleAndUrlSerializer(serializers.Serializer):
+    title = serializers.CharField()
```
ababic: We may have to use something dynamic here to source a value, as the 'title' value is typically blank in anything other than English.

core/serializers.py
ababic: > why isn't the home page of an app the parent of the other pages?

I'm working right now on changing this for international and domestic. I'm hoping for the changes to be merged today.

ababic: > why isn't the home page of an app the parent of the other pages?

I'm working right now on changing this for international and domestic. I'm hoping for the changes to be merged today.

core/models.py
MichaelWalker: yeah I deliberately didn't suggest anything because I couldn't think of a good suggestion.
`ancestor_content_pages` was the least bad I came up with, but I don't love that either! :)

core/serializers.py
eadpearce: you mean if the translations aren't populated? then we wouldn't show that page anyway right?

ababic: No, the panels are configured in a way that doesn't allow editors to specify translated values for `title`, so even if the page has been translated we won't get a translated title this way. The fields that are translated vary from page to page, but I've narrowed them down to the following in international:

- `article_title`
- `landing_page_title`
- `campaign_heading`
- `display_title`
- `heading`
- `hero_title`

Some pages also have a `breadcrumbs_label` field, which might even take precedence here.

eadpearce: ohhhh I see now. couldn't we just edit the panels to make `title_[language code]` editable?

ababic: That would be a better fix, but not 100% sure on the side-effects of that. There are a lot of panel sets to update, that's for sure :)

core/models.py
ababic: @eadpearce Any thoughts on the name `ancestors_from_app_root()`?

eadpearce: oh maybe `ancestors_in_app`?

ababic: Yep. works for me! :)

PR #490, opened 2019-06-11
Changed field types and max length for project details section in opp…

jessgilbert:
…ortunity page


great_international/migrations/0034_auto_20190611_1113.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('great_international', '0033_auto_20190611_1057'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
MichaelWalker: Is this clever enough to retain all of the data we have, or will we loose the existing values?

(Do we mind?)

jessgilbert: I would expect it to be fine as it is altering an existing field and there has been no name change, but I could be wrong :/

jessgilbert: Checked it, I changed a CharField to a MarkDown and the data didn't get erased so looks like it will be fine to me :) Although codecov/patch isn't finishing not sure why? It has been a while

ababic:
Can I request that you hold off merging this until #485 is merged? There are already 2 merge commits in that PR due to having to reconcile the migration trees in international. I'll be finished very soon.

ababic:
Hi @jessgilbert, I think you may need to create a merge commit to join up these migrations and the new ones from `develop`. The following should do it:
```
make debug_manage cmd="makemigrations great_international --merge"
```

PR #501, opened 2019-06-13
[CMS 1507] Upgrade Django to 2.2 and Wagtail to 2.5

ababic:
The bulk of these changes actually come from updating `directory-constants` to the latest version as there a has been a lot of moving/renaming of constant values between the previous and latest versions. There were actually very few changes needed to make the codebase compatible with later versions of Django (some url changes, using `MIDDLEWARE` instead of `MIDDLEWARE_CLASSES`, updating custom middleware to use Django's `MiddlewareMixin`).

From a dependency perspective, there were quite a few third-party apps that needed updating to support Django 2.2. Where possible I've upgraded to the latest versions to benefit from any bug fixes or performance improvements that may have been made. Probably the most important change here is the upgrade of `wagtail-modeltranslation` to the latest version (0.10.2) from an old pinned commit. But, despite this big jump, things seem to be working as they did before (no test changes or migrations were needed).

For the dependencies within our control (`sigauth`, `directory-healthcheck`, `directory-components`, `directory-constants`) new releases were needed, but the changes there were very simple.

For anything not self-explanatory, please see comments.


activitystream/tests/test_views.py
```diff
@@ -8,6 +8,7 @@
 from rest_framework.reverse import reverse
 from rest_framework.test import APIClient
 from django.conf import settings
+from django.utils import timezone
```
ababic: These changes weren't necessary, but result in much cleaner test output (many `UserWarning`s appear otherwise)

conf/urls.py
```diff
@@ -72,13 +72,11 @@
 urlpatterns = [
```
ababic: From Django 2.0, `include()` and URL namespace declaration works a little differently. See about half way down this list: https://docs.djangoproject.com/en/2.2/releases/2.0/#features-removed-in-2-0

conf/wsgi.py
```diff
@@ -10,9 +10,7 @@
 import os
 
 from django.core.wsgi import get_wsgi_application
-from whitenoise.django import DjangoWhiteNoise
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "conf.settings")
 
 application = get_wsgi_application()
-application = DjangoWhiteNoise(application)
```
ababic: From [version 4.0 ](http://whitenoise.evans.io/en/stable/changelog.html#v4-0) WhiteNoise works entirely using Middleware. We're now using 4.1.2.

core/forms.py
```diff
@@ -19,12 +19,6 @@ class CopyToEnvironmentForm(forms.Form):
 
 class WagtailAdminPageForm(WagtailAdminPageForm):
 
-    @property
-    def media(self):
-        media = super().media
-        media.add_js(['core/js/sum_required_localised_fields.js'])
-        return media
```
ababic: From Django 2.0, this was preventing any of the form js being included in the page. This override has been replaced with [`core.wagtail_hooks. add_sum_required_fields_js()`](https://github.com/uktrade/directory-cms/blob/3f4250e413cbdb182ea282e87d33dcf6c524909b/core/wagtail_hooks.py#L82), which uses Wagtail's `insert_editor_js` hook to add the JS instead.

core/tests/test_views.py
```diff
@@ -270,7 +271,7 @@ def test_add_page_prepopulate(
             translated_fas_industry_page.slug
         )
 
-    response = admin_client.post(url, post_data)
+    response = admin_client.post(url, clean_post_data(post_data))
```
ababic: From [Django 2.2](https://docs.djangoproject.com/en/2.2/releases/2.2/#miscellaneous), the test client raises a TypeError when `None` values are supplied for post values. A `clean_post_data()` method is used here to remove any such values beforehand.

core/tests/test_views.py
```diff
@@ -303,12 +304,13 @@ def test_add_page_prepopulate_missing_content_type(
...
-        'hero_image',
-    ])
-    response = admin_client.post(url, {**data, **cluster_data})
+    post_data = model_to_dict(
+        translated_fas_industry_page,
+        exclude=['go_live_at', 'expire_at', 'hero_image']
+    )
+    post_data.update(cluster_data)
+
+    response = admin_client.post(url, clean_post_data(post_data))
```
ababic: From [Django 2.2](https://docs.djangoproject.com/en/2.2/releases/2.2/#miscellaneous), the test client raises a TypeError when `None` values are supplied for post values. A `clean_post_data()` method is used here to remove any such values beforehand.

core/widgets.py
```diff
@@ -45,8 +45,8 @@ def __init__(self, *args, select2_options={}, **kwargs):
         super().__init__(*args, **kwargs)
         self.select2_options = select2_options
 
-    def render(self, name, value, attrs=None):
-        html = super().render(name, value, attrs)
+    def render(self, name, value, attrs=None, renderer=None):
```
ababic: From [Django 2.1](https://docs.djangoproject.com/en/2.2/releases/2.1/#features-removed-in-2-1) the render() method of widgets must accept a `render` argument

groups/models.py
```diff
@@ -31,7 +31,9 @@ class GroupInfo(models.Model):
         (5, '5 (highest)'),
     ]
 
-    group = models.OneToOneField('auth.Group', related_name='info')
+    group = models.OneToOneField(
+        'auth.Group', related_name='info', on_delete=models.CASCADE
```
ababic: From [`Django 2.0`](https://docs.djangoproject.com/en/2.2/releases/2.0/#features-removed-in-2-0) the `on_delete` argument is required for `ForeignKey` and `OneToOneField` in code and migrations.

users/middleware.py
```diff
@@ -15,7 +16,7 @@ def process_request(self, request):
 
         user = request.user
 
-        if not user.is_authenticated() or user.is_superuser:
+        if not user.is_authenticated or user.is_superuser:
```
ababic: From [Django 2.0](https://docs.djangoproject.com/en/2.2/releases/2.0/#features-removed-in-2-0) `is_authenticated` works as a property only

core/tests/test_views.py
```diff
@@ -250,10 +251,10 @@ def test_add_page_prepopulate(
...
     post_data = {
         **model_as_dict,
-        'hero_image': image.file.name,
-        'introduction_column_one_icon': image.file.name,
-        'introduction_column_two_icon': image.file.name,
-        'introduction_column_three_icon': image.file.name,
+        'hero_image': image.pk,
+        'introduction_column_one_icon': image.pk,
+        'introduction_column_two_icon': image.pk,
+        'introduction_column_three_icon': image.pk,
```
ababic: From Django 2.2 posting image filenames to this view instead of ids causes a `ValueError` to be raised when cleaning the form. Is this functionality is important to preserve? Additional work may be required here, but I couldn't actually find where in the view/form code this conversion is taking place?

users/forms.py
```diff
@@ -65,7 +65,7 @@ def get_groups_queryset(self, user):
                 .distinct()
             )
             return Group.objects.filter(
-                page_permissions__page_id=entry_points_ids
+                page_permissions__page_id__in=entry_points_ids
```
ababic: I couldn't pin this down to a specific item in the release notes, but from Django 2.0, this filter was triggering an exception, which was fixed by this change

core/views.py
```diff
@@ -340,25 +359,12 @@ def get_parent(self):
...
-        form = form_class(
-            data=form.data,
-            instance=self.page,
-            parent_page=parent_page,
-        )
-        edit_handler = edit_handler.bind_to_instance(
-            instance=self.page,
-            form=form,
-            request=self.request
-        )
```
richtier: nice to see how this is simplified :) good stuff.

does copy upstream work when tested manually?

ababic: @richtier I haven't been able to figure out why yet, but I cannot get this view to work locally when posting to it from dev or staging. For some unknown reason, I get redirected to the login page ever time, which redirects me back to this view as a GET request (which correctly fails with a 405, because the POST data has gone). I don't suppose you have ever witnessed this behaviour locally? It seems to happen even if I revert the view changes in the module.

richtier: if you go to stage directly, do are you able to log in?
once logged in, are you still logged in once you try coping upstream?

PR #504, opened 2019-06-17
CMS-1624 Move FAS homepage and contact in international

sdonk:


great_international/apps.py
```diff
@@ -26,6 +26,8 @@ def ready(self):
         cache.InvestHighPotentialOpportunityDetailPageSubscriber.subscribe()
         cache.InvestHighPotentialOpportunityFormPageSubscriber.subscribe()
         cache.InvestHighPotentialOpportunityFormSuccessPageSubscriber.subscribe()  # noqa
+        cache.InternationalTradeHomePageSubscriber.subscribe()
```
MichaelWalker: I don't think the word 'International' adds anything here. Is there anything to stop us just having `Trade...`?

makefile
```diff
@@ -144,8 +144,6 @@ update_db_template: \
 	debug_migrate
 	pg_dump \
 		--no-owner \
-		--exclude-table=auth_user \
-		--exclude-table=users_userprofile \
 		--file=db_template.sql \
```
richtier: what about excluding the table data?

makefile
```diff
@@ -144,8 +144,6 @@ update_db_template: \
 	debug_migrate
 	pg_dump \
 		--no-owner \
-		--exclude-table=auth_user \
-		--exclude-table=users_userprofile \
```
MichaelWalker: Was deleting these lines deliberate? They were added quite recently and seemed like a sensible addition!

makefile
sdonk: ops

makefile
sdonk: They produce an unusable template. `--exclude-table-data` is the right flag to use

PR #512, opened 2019-06-21
Update version of django-debug-toolbar to fix django 2 incompatibility.

MichaelWalker:


requirements_test.txt
```diff
@@ -23,7 +23,7 @@ directory-constants==17.1.0
 directory-healthcheck==1.1.1
 dj-database-url==0.5.0
 django-admin-ip-restrictor==1.0.0
-django-debug-toolbar==1.11
```
MichaelWalker: It isn't really a test requirement.

The feature is turned off in production via an env-var on all environments, but in principal we might want to turn it on on dev or staging, and at the moment we wouldn't be able to, as the requirement would be missing.

sdonk: I think it was in requirements_test because we only wanted it on local machines. We never had the necessity of enabling it on dev or staging and I'm not sure we will want that. @richtier ?

ababic: Ideally, I think we'd have a separate dependency list for development - but I don't think it's worth the effort to change that here. I would suggest leaving things here for now, as it's one less thing to install/update during deployment.

richtier: agree we should keep it in `requirements_test`. `_test` is a misnomer: it's really "non-prod" - so test and dev munged together.

MichaelWalker: I'd actually prefer to have a consistent dependency list, as it reduces the chance of surprising 'works on my machine' type issues and makes everything easier to reason about.

It sounds like I'm outvoted here though, so I'll put the toolbar back where I found it! 😊

PR #513, opened 2019-06-21
Turn off django-debug-toolbar

MichaelWalker:
Prevents creating of 'InternationalHomePage'.
There should be a fix for this in Wagtail 2.6, so it can be turned back on then.

If we need it in the short term, then the following config might be helpful - https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#debug-toolbar-panels


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 ### Fixed Bugs
 
 - CMS-1678 - Fix SSO login bug by using a later version of django-staff-sso-client
-
+- no ticket - Turn off Debug Toolbar locally by default, as it prevents some pages from being created (fix in Wagtail 2.6)
```
ababic: @MichaelWalker tiny thing, but I think we might want to keep the blank line below this

ababic:
I think we can safely ignore codecov in this instance.

PR #520, opened 2019-06-25
CMS-1676 Force json loads

sdonk:


core/upstream_serializers.py
```diff
@@ -99,6 +98,7 @@ def serialize_value(cls, value):
 
     @classmethod
     def deserialize_value(cls, value):
+        value = helpers.dictify_data(value)
```
sdonk: this is where it explodes according to sentry

PR #525, opened 2019-06-26
Marketing Article Page Model

erdem:
https://uktrade.atlassian.net/browse/XOT-922


export_readiness/translation.py
```diff
@@ -65,6 +65,7 @@ class ArticleListingPageTranslationOptions(BaseTranslationOptions):
 
 
 @register(models.ArticlePage)
+@register(models.MarketingArticlePage)
 class ArticlePageTranslationOptions(BaseTranslationOptions):
```
richtier: please add 
```
class MarketingArticlePageTranslationOptions(BaseTranslationOptions):
    fields = []
```

export_readiness/cache.py
```diff
@@ -61,6 +61,19 @@ class ArticlePageSubscriber(AbstractDatabaseCacheSubscriber):
...
+    model = models.MarketingArticlePage
+    subscriptions = [
+        models.ArticlePage,
+        models.TopicLandingPage,
+        models.SuperregionPage,
+        models.CountryGuidePage,
+        models.CampaignPage,
+        models.ArticleListingPage,
+        models.Tag,
+    ]
```
richtier: `MarketingArticlePage` needs to only subscribe to pages that are included in it's serializer (because the serialized data is stored in the cache). As far as I can tell `TopicLandingPage`, `SuperregionPage`, `CountryGuidePage`, `CampaignPage`, and `ArticleListingPage` are not included in the serializer.

PR #537, opened 2019-07-02
CI-269 Categorise articles by article type and sector

Kenneth-Siu:


great_international/models.py
```diff
@@ -1477,7 +1530,12 @@ class InternationalArticlePage(BaseInternationalPage):
         FieldPanel('title_en_gb'),
         FieldPanel('slug'),
         FieldPanel('uses_tree_based_routing'),
-        FieldPanel('tags', widget=CheckboxSelectMultiple)
+        FieldPanel('article_types', widget=CheckboxSelectMultiple),
+        FieldPanel('tags', widget=CheckboxSelectMultiple),
```
MichaelWalker: We don't seem to have a class variable for this - is that right?

MichaelWalker: Also, do we need some updates to the serializers to handle the new fields?

PR #540, opened 2019-07-09
Removed prioritised checkbox from opportunity page

jessgilbert:

 - [x] Changelog entry added.

This will break great-international-ui if merged without the front-end changes. Am closing this until the front-end change have been released, to avoid great-international-ui having an unnecessary dependence on the cms when releasing. Is now a clean up task in the backlog https://uktrade.atlassian.net/browse/CI-328


jessgilbert:
This will break `great-international-ui` if merged without the front-end changes. Am closing this until the front-end change have been released, to avoid `great-international-ui` having an unnecessary dependence on the cms when releasing. Is now a clean up task in the backlog https://uktrade.atlassian.net/browse/CI-328

PR #542, opened 2019-07-10
Feature/xot 932 add country market pages to search

WillTaylor22:


activitystream/serializers.py
```diff
@@ -4,19 +4,37 @@
 class ArticlePageSerializer(serializers.Serializer):
```
eadpearce: instead of having one serializer that does both pages i think it'd be better to have a separate `CountryGuidePageSerializer` and check which one the page needs in `ActivityStreamView`. otherwise as we add more page types to search this serializer may get very complicated

activitystream/serializers.py
```diff
@@ -4,19 +4,37 @@
...
-            'object': {
-                'type': 'Article',
-                'id': 'dit:cms:Article:' + str(obj.id),
-                'name': obj.article_title,
-                'summary': obj.article_teaser,
-                'content': obj.article_body_text,
-                'url': obj.get_url()
-            },
-        }
+        if obj.__class__.__name__ == 'ArticlePage':
```
richtier: this is brittle - if the model name changed then this would beak. different serializer :100: please

FWIW, if you wanted to do similar thing elsewhere, it would be more cromulent to do `if isinstance(obj, ArticlePage)`

db_template.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
```
sdonk: There are no models changes. Does this new template speed up tests?

activitystream/serializers.py
WillTaylor22: I might be missing something but the issue is that we are passing in a set of models of different types into the serializer, using many=True (i.e. PageSerializer(page_qs, many=True)). Therefore we don't know exactly what data will be provided. All we know is we have a queryset that we want to serialize, and for me, that's all the view should know.

Appreciate the isinstance() recommendation!

db_template.sql
WillTaylor22: Yes that's correct

activitystream/serializers.py
WillTaylor22: The issue is that we are passing in a set of models of different types into the serializer, using many=True (i.e. PageSerializer(page_qs, many=True)). Therefore we don't know exactly what data will be provided / serializer will be needed. Usually both are needed in the same query.

All we know in the view is we have a mixed queryset that we want to serialize, and for me, that's all the view should know.

However have changed it to make it a little cleaner!

eadpearce: for reference we do that here already https://github.com/uktrade/directory-cms/blob/d6a8117e9e2b206c450d6201c869a9d774ff887a/export_readiness/serializers.py#L55..#L74

db_template.sql
sdonk: It doesn't seem to be working. Some migrations are still running before the build 
https://circleci.com/gh/uktrade/directory-cms/5304

Considering the template doesn't seem to be doing anything I'd consider removing it

richtier: tests are now green so this seems resolved

sdonk: Nope. The point of this new template was for the test runner not to run migrations and speed up tests, it was never to fix tests.
Migrations are still running at the beginning of the test suite.

PR #543, opened 2019-07-11
CMS-1648 tree base routing helpers

sdonk:

 - [X] Changelog entry added.


core/management/commands/enable_tree_based_routing.py
```diff
@@ -0,0 +1,39 @@
...
+class Command(BaseCommand):
+    help = 'Enable the tree based routing flag on pages. ' \
+           'If --servicename is not specified, the flag is enabled on ' \
+           'ALL the pages.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--servicename',
+            help='Select international or domestic',
+            choices=['international', 'domestic']
```
richtier: https://docs.python.org/3/library/argparse.html#choices
> Any object that supports the in operator can be passed as the choices value, so dict objects, set objects, custom containers, etc. are all supported.
so I think you can pass `choices=SERVICES_MAPPING` and not do the dict lookup in `handle`

core/management/commands/enable_tree_based_routing.py
```diff
@@ -0,0 +1,39 @@
...
+        queryset = Page.objects.exclude(title__isnull=True)
+        if options['servicename']:
+            service_name = SERVICES_MAPPING[options['servicename']]
+            queryset = filters.ServiceNameFilter().filter_service_name(
+                queryset=queryset,
+                name=None,
+                value=service_name,
+            )
+        for page in queryset:
+            page.specific.uses_tree_based_routing = True
```
richtier: is `page.specific` cached? I have a suspicion that we need to do
```
page = page.specific
# do some mutation
page.save()
```

alternatively we can do `for page in queryset.specific():` and do it in one place

great_international/management/commands/set_missing_translated_titles.py
```diff
@@ -37,11 +37,11 @@ def handle(self, *args, **options):
             value=cms.GREAT_INTERNATIONAL,
         )
         for page in queryset.specific():
-            print('--------------------------------------------------------')
-            print("{}: {} (ID:{})".format(
+            self.stdout.write('--------------------------------------------------------')
+            self.stdout.write('{}: {} (ID:{})'.format(
```
richtier: not a fan of f strings?
```
self.stdout.write(f'{page.__class__.__name__}: {page} (ID:{page.id})'
```

core/management/commands/enable_tree_based_routing.py
sdonk: I tried `queryset.specific()` but it tries to call `Page` and the property `uses_tree_based_routing` doesn't exist on that class

great_international/management/commands/set_missing_translated_titles.py
sdonk: I just replaced Andy's `print()` with `self.stdout.write`, as per Django's documentation.
I'll add the f strings

core/management/commands/enable_tree_based_routing.py
sdonk: I can do `choices=SERVICES_MAPPING`and keep the lookup. If you pass a dict then either the keys or the values can be passed in the command, but it's easier to remember `domestic` than `EXPORT_READINESS` when you use the command

PR #545, opened 2019-07-12
Ci 324 capital invest sub sectors

jessgilbert:

 - [x] Changelog entry added.


great_international/models/great_international.py
```diff
@@ -23,12 +23,12 @@
 from .base import BaseInternationalPage
 
 
-class InternationalSectorPage(
+class AbstractInternationalSectorPage(
```
MichaelWalker: I think we've generally used `Base...` in the past for classes which we extend but don't directly use.
Unless there is a good reason to use `Abstract...` instead this time we should probably stick to the convension.

PR #547, opened 2019-07-12
CI-322 Add about dit services pages

Kenneth-Siu:

 - [x] Changelog entry added.

Updating db_template.sql isn't playing nice on my machine. Should be okay for one migration.


Kenneth-Siu:
Updating db_template.sql isn't playing nice on my machine. Should be okay for one migration.

great_international/serializers.py
```diff
@@ -1666,3 +1666,46 @@ def get_industry_options(self, instance):
...
+
+class AboutDitLandingPage(BasePageSerializer):
+    breadcrumbs_label = serializers.CharField()
+    hero_title = serializers.CharField()
+    hero_image = wagtail_fields.ImageRenditionField('original')
+
+
+class AboutDitServiceFieldSerializer(serializers.Serializer):
+    icon = wagtail_fields.ImageRenditionField(
+        'original'
```
MichaelWalker: Do we not want a specific image size for the icons?

great_international/serializers.py
```diff
@@ -1666,3 +1666,46 @@ def get_industry_options(self, instance):
...
+    link_url = serializers.CharField()
+
+
+class AboutDitServicesPageSerializer(BasePageSerializer):
+    breadcrumbs_label = serializers.CharField()
+    hero_title = serializers.CharField()
+    hero_image = wagtail_fields.ImageRenditionField('original')
+    teaser = core_fields.MarkdownToHTMLField()
+    teaser_image = wagtail_fields.ImageRenditionField('fill-640x360')
+    case_study_image = wagtail_fields.ImageRenditionField('original')
```
MichaelWalker: likewise I think I'd expect some forced sizes on these images.

great_international/serializers.py
Kenneth-Siu: We haven't generally done so for icons. The serializer section is for serving up a formatted version of the image, which we don't want to do for the icons since we're assuming they're saved at the correct size (or if not resized in the CSS)

great_international/serializers.py
Kenneth-Siu: Equally, we haven't resized the case study images because, like the hero images, they stretch over the whole width of the screen.

PR #548, opened 2019-07-12
Update invest homepage model in great international

eadpearce:

 - [x] Changelog entry added.


great_international/panels/invest.py
```diff
@@ -172,6 +172,7 @@ class InvestInternationalHomePagePanels:
     settings_panels = [
         FieldPanel('title_en_gb'),
         FieldPanel('slug'),
+        FieldPanel('uses_tree_based_routing'),
```
sdonk: We don't need this anymore. Every page in international will have it enabled by default

PR #551, opened 2019-07-17
Ci 324 capital invest sub sectors

jessgilbert:
No changelog as this is adding to a ticket and bulk change already on there from pr before

### DIFF
--- a/great_international/serializers.py
+++ b/great_international/serializers.py
@@ -1136,28 +1136,21 @@ class CapitalInvestOpportunityListingSerializer(BasePageSerializer):
 
     sector_with_sub_sectors = serializers.SerializerMethodField()
 
+    def get_sub_sector_headings(self, sector):
+        return [sub_sector['heading'] for sub_sector
+                in sector['child_sub_sectors']]
+
     def get_sector_with_sub_sectors(self, instance):
-        sector_with_sub_sectors = {}
+
         all_sectors = InternationalSectorPage.objects.live().public()
-
-        if not all_sectors:
-            return sector_with_sub_sectors
-
-        serialized_sectors = InternationalSectorPageSerializer(
+        sectors = InternationalSectorPageSerializer(
             all_sectors,
             many=True,
             allow_null=True,
             context=self.context
-        )
-        for sector in serialized_sectors.data:
-            if 'child_sub_sectors' in sector:
-                key = sector['heading']
-                sector_with_sub_sectors[key] = []
-                for sub_sector in sector['child_sub_sectors']:
-                    value = sub_sector['heading']
-                    sector_with_sub_sectors[key].append(value)
+        ).data
 
-        return sector_with_sub_sectors
+        return {sector['heading']: self.get_sub_sector_headings(sector) for sector in sectors}
 
 
 class CapitalInvestOpportunityPageSerializer(
--- a/great_international/serializers.py
+++ b/great_international/serializers.py
@@ -1134,28 +1134,30 @@ class CapitalInvestOpportunityListingSerializer(BasePageSerializer):
         )
         return serializer.data
 
-    sub_sector_with_parent = serializers.SerializerMethodField()
+    sector_with_sub_sectors = serializers.SerializerMethodField()
 
-    def get_sub_sector_with_parent(self, instance):
-        sector_with_parent = {}
-        queryset = InternationalSubSectorPage.objects.live().public()
+    def get_sector_with_sub_sectors(self, instance):
+        sector_with_sub_sectors = {}
+        all_sub_sectors = InternationalSubSectorPage.objects.live().public()
+        all_sectors = InternationalSectorPage.objects.live().public()
 
-        if not queryset:
+        if not all_sub_sectors:
             return []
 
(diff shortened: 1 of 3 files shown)


great_international/serializers.py
```diff
@@ -1120,6 +1134,31 @@ def get_opportunity_list(self, instance):
...
 
+    sector_with_sub_sectors = serializers.SerializerMethodField()
+
+    def get_sector_with_sub_sectors(self, instance):
+        sector_with_sub_sectors = {}
+        all_sub_sectors = InternationalSubSectorPage.objects.live().public()
+        all_sectors = InternationalSectorPage.objects.live().public()
+
+        if not all_sub_sectors:
+            return []
```
richtier: could you return an empty dict here to avoid the function sometimes returning a list and sometimes returning a dict :)

great_international/serializers.py
```diff
@@ -1120,6 +1134,31 @@ def get_opportunity_list(self, instance):
...
+
+        serializer = InternationalSubSectorPageSerializer(
+            all_sub_sectors,
+            many=True,
+            allow_null=True,
+            context=self.context
+        )
+        for sector in all_sectors:
+            sector_with_sub_sectors[sector.heading] = []
```
richtier: [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) would make this a little easier
```
from collections import defaultdict

sector_with_sub_sectors = defaultdict(list)
```
that shoild avoid the no need to do 
```
for sector in all_sectors:
    sector_with_sub_sectors[sector.heading] = []
```

...I think

great_international/serializers.py
```diff
@@ -1120,6 +1134,31 @@ def get_opportunity_list(self, instance):
...
+
+        for sub_sector in serializer.data:
+            sector_with_sub_sectors[sub_sector['parent']['heading']]\
+                .append(sub_sector['heading'])
```
richtier: bit hard to read. 
```
key = sub_sector['parent']['heading']
sector_with_sub_sectors[key].append(sub_sector['heading'])
```
would be a bit easier

great_international/serializers.py
```diff
@@ -573,6 +573,20 @@ def get_related_opportunities(self, instance):
...
 
 
+class InternationalSubSectorPageSerializer(
+    BaseInternationalSectorPageSerializer,
+    ParentPageSerializerHelper
+):
+    parent = serializers.SerializerMethodField()
+
+    def get_parent(self, obj):
+        sector_parent = self.get_parent_page_data_for(
```
MichaelWalker: (not a big deal but...) the assignment to `sector_parent` doesn't really add anything here.
I'd just 

```python
return self.get_parent_page_data_for()
```

great_international/serializers.py
```diff
@@ -1120,6 +1134,31 @@ def get_opportunity_list(self, instance):
...
+            context=self.context
+        )
+        for sector in all_sectors:
+            sector_with_sub_sectors[sector.heading] = []
+
+        for sub_sector in serializer.data:
+            sector_with_sub_sectors[sub_sector['parent']['heading']]\
+                .append(sub_sector['heading'])
+
+        return sector_with_sub_sectors
```
MichaelWalker: One other way we could approach this (which _might_ end up being nicer) is to make it so that an `SectorPage` serializes it's children 'sub-sector' pages.

That would then make this method really simple, at the cost of adding a bit of complexity to the `SectorPage` serializer.

That's probably a net neutral change right now, but it might make things easier going forward.
For example, if we think that a sector page will soon want to display its sub_sectors (I think this is quite likely) then we can use the 'children' serialization for both things.

What do you think?

jessgilbert: Agreed, will do that :)

great_international/serializers.py
```diff
@@ -1138,24 +1138,24 @@ def get_opportunity_list(self, instance):
...
-        for sub_sector in serializer.data:
-            sector_with_sub_sectors[sub_sector['parent']['heading']]\
-                .append(sub_sector['heading'])
+        for sector in serialized_sectors.data:
+            if 'child_sub_sectors' in sector:
+                key = sector['heading']
+                sector_with_sub_sectors[key] = []
+                for sub_sector in sector['child_sub_sectors']:
+                    value = sub_sector['heading']
+                    sector_with_sub_sectors[key].append(value)
```
MichaelWalker: This looks right, but I wonder if we can keep making it a little more readable. How about

```python
def get_sub_sector_titles(self, sector):
    return [sub_sector['heading'] for sub_sector in sector['child_sub_sectors']]

def get_sector_with_sub_sectors(self, instance):
    all_sectors = InternationalSectorPage.objects.live().public()
    
    if not all_sectors:
        return {}

    serialized_sectors = InternationalSectorPageSerializer(
	    all_sectors,
            many=True,	
            allow_null=True
            context=self.context
    ).data
    
    return {sector['heading'], self.get_sub_sector_titles(sector) for sector in serialized_sectors}
```

jessgilbert: 👍 Done

PR #564, opened 2019-07-23
Tidy up unused models and translations

eadpearce:

 - [ ] Changelog entry added.
 - [ ] (if adding new models) Add an updated version of db_template.sql
 - [ ] (if adding HTML) Change is accessible to the standards we subscribe to.
 - [ ] (if changing the UI) Add a printscreen to the PR


eadpearce:
can't get migrations to work. going to abandon this for now

PR #565, opened 2019-07-23
CI-368 Add ebook promotional section to benefits page

Kenneth-Siu:
 - [x] Changelog entry added.


great_international/migrations/0053_auto_20190723_1622.py
```diff
@@ -0,0 +1,242 @@
...
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to='wagtailimages.Image'),
+        ),
+        migrations.AddField(
+            model_name='aboutukwhychoosetheukpage',
+            name='ebook_section_image_fr',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to='wagtailimages.Image'),
+        ),
+        migrations.AddField(
+            model_name='aboutukwhychoosetheukpage',
+            name='ebook_section_image_ja',
```
eadpearce: does this image field really need translations? it will cut down migration time significantly if it isn't translated

eadpearce:
please also remember to run `make update_db_template` to update the db template for the tests

great_international/migrations/0053_auto_20190723_1622.py
Kenneth-Siu: As per Edie comment on slack, needs alt text.

PR #566, opened 2019-07-23
Small HPO tweaks

eadpearce:

 - [x] Changelog entry added.

### DIFF
--- a/great_international/models/great_international.py
+++ b/great_international/models/great_international.py
@@ -9,7 +9,7 @@ from directory_constants import slugs
 from core.model_fields import MarkdownField
 
 from core.models import (
-    ExclusivePageMi xin,
+    ExclusivePageMixin,
     WagtailAdminExclusivePageMixin,
     FormPageMetaClass,
 )
--- a/great_international/models/great_international.py
+++ b/great_international/models/great_international.py
@@ -9,7 +9,7 @@ from directory_constants import slugs
 from core.model_fields import MarkdownField
 
 from core.models import (
-    ExclusivePageMixin,
+    ExclusivePageMi xin,
     WagtailAdminExclusivePageMixin,
     FormPageMetaClass,
 )
@@ -754,14 +754,13 @@ class InternationalHomePageOld(
     visit_uk_cta_text = models.CharField(max_length=255)
 
 
+# !!! TO BE REMOVED !!!
 class InternationalRegionPage(
     BaseInternationalPage,
     panels.InternationalRegionPagePanels,
 ):
     parent_page_types = ['great_international.InternationalHomePage']
-    subpage_types = [
-        'great_international.InternationalLocalisedFolderPage'
-    ]
+    subpage_types = []
 
     tags = ParentalManyToManyField(Tag, blank=True)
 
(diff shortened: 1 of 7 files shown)


great_international/models/invest.py
```diff
@@ -249,12 +249,13 @@ class InvestHighPotentialOpportunitiesPage(
...
     def save(self, *args, **kwargs):
-        self.title = self.get_verbose_name()
+        # title is used for tree_based_breadcrumbs that are displayed in the UI
+        self.title = 'High potential opportunities'
         return super().save(*args, **kwargs)
 
 
 class InvestHighPotentialOpportunityFormPage(
-    WagtailAdminExclusivePageMixin,
+    ExclusivePageMixin,
```
eadpearce: using `ExclusivePageMixin` to prevent the user from modifying the slug

great_international/models/invest.py
```diff
@@ -508,8 +509,9 @@ class InvestHighPotentialOpportunityDetailPage(
 
 
 class InvestHighPotentialOpportunityFormSuccessPage(
-    BaseInternationalPage,
     panels.InvestHighPotentialOpportunityFormSuccessPagePanels,
+    ExclusivePageMixin,
+    BaseInternationalPage,
```
eadpearce: inheritance order here was stopping the admin panels from showing up

great_international/serializers.py
```diff
@@ -1511,6 +1510,11 @@ def get_other_opportunities(self, instance):
         return serializer.data
 
 
+class InvestHighPotentialOpportunityListItemSerializer(BasePageSerializer):
```
eadpearce: no point serializing the whole page here since we only use these two fields on the FE

great_international/models/great_international.py
```diff
@@ -9,7 +9,7 @@
 from core.model_fields import MarkdownField
 
 from core.models import (
-    ExclusivePageMixin,
+    ExclusivePageMi xin,
```
richtier: `ExclusivePageMi xin` :)

great_international/models/invest.py
```diff
@@ -249,12 +249,13 @@ class InvestHighPotentialOpportunitiesPage(
     ]
 
     def save(self, *args, **kwargs):
-        self.title = self.get_verbose_name()
+        # title is used for tree_based_breadcrumbs that are displayed in the UI
+        self.title = 'High potential opportunities'
```
richtier: the cached value will not receive this change until the page is manually saved. This is just a reminder we will need to do that when we release to prod

great_international/serializers.py
```diff
@@ -1526,12 +1530,12 @@ class Meta:
 
     def get_opportunity_list(self, instance):
         queryset = (
-            HighPotentialOpportunityDetailPage.objects.all()
+            InvestHighPotentialOpportunityDetailPage.objects.all()
```
richtier: do we need to change the cache subscriber to subscribe to `InvestHighPotentialOpportunityDetailPage`?

great_international/models/great_international.py
eadpearce: LOL i have no idea how i managed to do this

great_international/models/invest.py
eadpearce: this is a new page that doesn't exist in prod yet so hopefully this won't be a problem

great_international/serializers.py
eadpearce: already done that. but that's reminded me i should have them all subscribe to each other so breadcrumbs on each page are updated properly

PR #572, opened 2019-07-24
TT-1689 move tests in tests folder

sdonk:

 - [x] Changelog entry added.


tests/conftest.py
```diff
@@ -19,9 +19,9 @@
 from core.models import RoutingSettings
 from groups.models import GroupInfo
 from users.models import UserProfile
-from great_international.tests.factories import InternationalHomePageFactory
-from find_a_supplier.tests.factories import IndustryPageFactory
-from users.tests.factories import UserFactory
+from .great_international.factories import InternationalHomePageFactory
+from .find_a_supplier.factories import IndustryPageFactory
+from .users.factories import UserFactory
```
MichaelWalker: I guess these must work otherwise the tests would fail...
Is the `.` doing something special or it the same as without it?

sdonk: It's looking into the same folder (`tests`). It's a shortcut for `from tests.great_international.factories import InternationalHomePageFactory`

eadpearce: it's just a relative import as opposed to an absolute import

MichaelWalker: Cool - I actually didn't know python had an explicit syntax for those! Nice!

(I thought it just scanned through its list of directories, and the current was always high enough on the list that everything works fine without the dot! Much nicer to be explicit about it though!)

PR #573, opened 2019-07-24
CI-366 Add CTA to UK benefits section of invest home page

Kenneth-Siu:
 - [x] Changelog entry added.
 - [x] ~~(if adding new models) Add an updated version of db_template.sql~~ Doesn't work on my machine


invest/models.py
```diff
@@ -713,6 +717,14 @@ class InvestHomePage(ExclusivePageMixin, ServiceHomepageMixin, BaseInvestPage):
 
         ),
 
+        MultiFieldPanel(
```
MichaelWalker: Is this multifield to make it collapsible?

Kenneth-Siu: Yep.

PR #576, opened 2019-07-29
Populated about dit landing page with fields

jessgilbert:

 - [x] Changelog entry added.
 - [x] (if changing the UI) Add a printscreen to the PR


great_international/models/great_international.py
```diff
@@ -1231,6 +1231,63 @@ class AboutDitLandingPage(panels.AboutDitLandingPagePanels, BaseInternationalPag
...
+        'wagtailimages.Image',
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+        blank=True
+    )
+
+    how_dit_help_title = models.CharField(max_length=255, blank=True)
+
+    related_page_one_image = models.ForeignKey(
```
eadpearce: are these related pages going to be external links? if not, you can add it as a related page like ```related_page_one = models.ForeignKey(
        'wagtailcore.Page',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+',
    )```

great_international/panels/great_international.py
```diff
@@ -1035,6 +1035,65 @@ class AboutDitLandingPagePanels:
...
+                HelpPanel('Required for section to show: Section one content'),
+                FieldPanel('section_one_content'),
+                ImageChooserPanel('section_one_image'),
+            ],
+        ),
+        MultiFieldPanel(
+            heading="Related pages",
+            classname='collapsible',
+            children=[
+                HelpPanel('Required for section to show: title and at least on related page'),
```
eadpearce: typo - "one related page"

great_international/models/great_international.py
jessgilbert: Yeah I thought about this, I need to check with UX designers probably, I could use related pages, but that then means adding a `featured_description` in those pages I am expecting to be added as at the minute we are expecting a link to, capital invest, trade and invest landing page so all these would need extra descriptions in their cms to be used for this card and are all different kind of pages and so have varying field names, so I couldn't decide which was a better option and so I went with this as this is more flexible, but then that does mean duplicating the image and title and url which you could get wrong. So I am open to suggestions on which to use 🤷‍♀

great_international/panels/great_international.py
jessgilbert: Thanks :D

great_international/models/great_international.py
jessgilbert: I prefer linking to a page with this as it's so much easier just clicking the right page and you haven't got to worry about getting the right image, the only reason I didn't do it is because I didn't know what happens if a page gets linked to that doesn't have a `featured_description` field that is in the related page serializer for example because any page could be clicked so it's likely one may have a different field name than the serializer is expecting, and if having to add the featured_description field to the pages I do want to link to is better than adding here?

jessgilbert: Oh! Or maybe I could add related pages but have a separate featured description field for each related page on this cms that is used, so that avoid it ever not being there from the related page or having to add it in the other pages?

jessgilbert: Turns out I entirely misunderstood the design thinking it goes to any page possibly and created all these issues with varying field names, but Kenny has just enlightened me the pages it is relating too is actually this landing pages children pages 😂 so all the problems are solved and I will just do 3 normal related pages :D (Doing it so you can pick 3 of the children rather than getting all children for now because I don't know if we want all or just to pick 3 - this may change)

eadpearce: the description we use for all our related page fields should probably use the page meta description which we should have anyway for SEO reasons. just have to make sure content editors know it's important to fill it out. this would solve a lot of our current problems with varying field names on pages

great_international/serializers.py
```diff
@@ -260,6 +260,41 @@ def icon(self):
         )
 
 
+class RelatedPagesCardsProxyDataWrapper:
```
eadpearce: we have a lot of serializers that do a similar thing to this. may be worth creating something reusable at some point to cut down on duplicated code

PR #582, opened 2019-07-30
Fix missing pages from children list

sdonk:

 - [x] Changelog entry added.


great_international/models/invest.py
```diff
@@ -319,6 +321,7 @@ class InvestHighPotentialOpportunityFormPage(
 
     slug_identity = 'contact'
     subpage_types = ['InvestHighPotentialOpportunityFormSuccessPage']
+    parent_page_types = ['InvestHighPotentialOpportunitiesPage']
```
eadpearce: this feels a bit belt-and-braces but hey if it still works i don't mind

PR #583, opened 2019-07-31
Ci 404 invest cms clean up

jessgilbert:

 - [x] Changelog entry added.


great_international/migrations/0062_auto_20190731_1236.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('great_international', '0061_merge_20190731_1124'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
richtier: is this backwards compatible or will removing this require frontend changes?

jessgilbert: Nope none of the fields I have removed are used anymore on master, no front-end change needed.

jessgilbert: The capital invest / set up / isd fields were the 3 cards which are now `featured_card_num` and that change has been released so are not used anywhere

PR #588, opened 2019-08-05
Ci 321 about uk landing page

jessgilbert:

 - [x] Changelog entry added.


tests/great_international/test_serializers.py
```diff
@@ -1011,3 +1011,52 @@ def test_opportunity_page_null_case_gets_opportunities_with_same_sector(
...
+@pytest.mark.django_db
+def test_about_uk_landing_page_has_all_sectors(
+        rf, international_root_page
+):
+    sector = InternationalSectorPageFactory(
+        parent=international_root_page,
+        slug='sector-one',
+    )
+
+    BaseInternationalSectorPageSerializer(
```
MichaelWalker: Do we need this? We don't seem to use the result of it at all.

great_international/serializers.py
```diff
@@ -300,6 +328,12 @@ class HowWeHelpSerializer(serializers.Serializer):
     icon = wagtail_fields.ImageRenditionField('original')
 
 
+class HowWeHelpWithTitleSerializer(serializers.Serializer):
```
MichaelWalker: I'm not sure I understand the name of this class (perhaps I'm just missing something). What does the `WithTitle` bit mean?

eadpearce: i believe it's to contrast with `HowWeHelpProxyDataWrapper` which only serializes icon and text and not title

great_international/translation.py
```diff
@@ -895,6 +895,60 @@ class AboutUkLandingPageTranslationOptions(
         'breadcrumbs_label',
         'hero_title',
         'hero_image',
+
+        'intro',
+
+        'why_choose_uk_title',
+        'why_choose_uk_content',
+        'why_choose_uk_image',
```
eadpearce: do the images require translation?

great_international/cache.py
```diff
@@ -16,7 +16,8 @@ class InternationalSectorPageSubscriber(AbstractDatabaseCacheSubscriber):
         capital_invest.CapitalInvestOpportunityPage,
         capital_invest.CapitalInvestOpportunityListingPage,
         great_international.InternationalSubSectorPage,
-        great_international.InternationalCuratedTopicLandingPage
+        great_international.InternationalCuratedTopicLandingPage,
+        great_international.AboutUkLandingPage
```
eadpearce: `AboutUkLandingPage` should also subscribe to `InternationalSectorPage` since it serializes sectors as child pages

great_international/serializers.py
jessgilbert: Yes indeed, the normal HowWeHelp has icon and text and no title

great_international/translation.py
jessgilbert: umm I didn't think you could translate only some of the fields? I get an error if I delete them, am I missing something?

tests/great_international/test_serializers.py
jessgilbert: Nope, removed 👍

great_international/cache.py
jessgilbert: 👍

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@
 - CI-328 - Removed unused fields for capital invest opportunity page
 - No-ticket - Fix models hierarchy
 - CI-404 - Clean up new Invest International home page fields that are no longer used
-
+- CI-321 - About UK landing page
```
sdonk: Git gets confused with the diff, this is now in proposed enhancements and not in the release section. @jessgilbert can you please push a fix?

great_international/translation.py
eadpearce: downside is you need to put them in a separate tab in the admin panels or `make_translated_interface` will try to look for fields that aren't there. e.g. `InvestInternationalHomePagePanels` has a separate tab for images in the admin

CHANGELOG.md
jessgilbert: This should be in proposed enhancements shouldn't it? It hasn't been released?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
+- CI-321 - About UK landing page
```
jessgilbert: @sdonk is this what you meant? It going funny once I had updated the branch? :)

great_international/translation.py
jessgilbert: Ah I see ok cool will do that :)

CHANGELOG.md
sdonk: Yep. It was under the release section!

PR #600, opened 2019-08-08
Ci 276 capital invest contact form

jessgilbert:

 - [x] Changelog entry added.


MichaelWalker:
Could you add a changelog?

great_international/models/capital_invest.py
```diff
@@ -686,3 +692,31 @@ class CapitalInvestOpportunityPage(
 
     contact_title = models.CharField(max_length=255, blank=True)
     contact_text = MarkdownField(blank=True)
+
+
+class CapitalInvestContactFormPage(
+    panels.CapitalInvestContactFormPagePanels, WagtailAdminExclusivePageMixin, BaseInternationalPage
+):
+    parent_page_types = ['great_international.InternationalCapitalInvestLandingPage']
+    slug_identity = 'contact'
```
MichaelWalker: do we have to worry about name clashes here, or is it safely scoped to its parent page?

(ie would it cause an issue if we created a different form in a different place with the same slug?)

great_international/models/capital_invest.py
```diff
@@ -686,3 +692,31 @@ class CapitalInvestOpportunityPage(
...
+
+
+class CapitalInvestContactFormSuccessPage(
+    panels.CapitalInvestContactFormSuccessPagePanels, WagtailAdminExclusivePageMixin, BaseInternationalPage
+):
+    parent_page_types = ['great_international.CapitalInvestContactFormPage']
+    slug_identity = 'success'
+
+    large_text = models.CharField(max_length=255)
+    small_text = MarkdownField(blank=True)
```
MichaelWalker: Admittedly I don't know what this page looks like, so perhaps this is as good as we can do.... but `large_text` and `small_text` seem quite vague.

Is there anything a bit more descriptive that would still make sense?

jessgilbert: That's a hard one :/ This is the Invest success page which I was going off of
image

the designers said I shouldn't have a hero on either page, so I have removed that and added large and small text for the cms to keep it flexible as I have nothing else to go off of at the moment.

great_international/models/capital_invest.py
jessgilbert: Hmm.. unsure

eadpearce: Slugs aren't required to be unique under apps anymore so this shouldn't be an issue. Though I think copy upstream still thinks they should be unique (but we need to fix that anyway)

eadpearce: Actually, we have this slug in constants so you can use `slugs.CONTACT_FORM_SLUG`

great_international/models/capital_invest.py
```diff
@@ -686,3 +692,31 @@ class CapitalInvestOpportunityPage(
...
+        return [
+            CapitalInvestContactFormSuccessPage
+        ]
+
+
+class CapitalInvestContactFormSuccessPage(
+    panels.CapitalInvestContactFormSuccessPagePanels, WagtailAdminExclusivePageMixin, BaseInternationalPage
+):
+    parent_page_types = ['great_international.CapitalInvestContactFormPage']
+    slug_identity = 'success'
```
eadpearce: Same here. You can use `slugs.FORM_SUCCESS_SLUG` from constants

great_international/models/capital_invest.py
eadpearce: Those two fields could potentially just be one markdown field

great_international/cache.py
```diff
@@ -239,3 +239,17 @@ class AboutUkWhyChooseTheUkPageSubscriber(
...
+):
+    model = capital_invest.CapitalInvestContactFormPage
+    subscriptions = []
+
+
+class CapitalInvestContactFormSuccessPageSubscriber(
+    AbstractDatabaseCacheSubscriber
+):
+    model = capital_invest.CapitalInvestContactFormSuccessPage
+    subscriptions = []
```
eadpearce: if `tree_based_breadcrumbs` are going to be displayed on these pages in the UI they'll need to subscribe to their parent pages

great_international/models/capital_invest.py
jessgilbert: This is a form and so I needed it to use it's own view so have given the specific url to go to the right view and doesn't use tree based routing. The forms I have seen use `CMSPageFromSlugMixin` which I have done, which uses look up by slug so having a non specific slug does matter in this case and so I am changing it to `capital-invest-contact`

eadpearce: If the page doesn't use tree based routing will the "view live" buttons in the cms still go to the right place?

jessgilbert: Yeah it does, its a child of the capital invest page so the view live goes to the right thing but when getting the cms response it uses look up by slug

eadpearce: What's the reason for looking up by slug rather than path then? Sorry I'm a bit confused!

PR #601, opened 2019-08-12
CMS-1792 Allow international article pages as child pages of international sectors

sdonk:
 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -657,6 +657,7 @@ class InternationalSectorPageSerializer(
     ChildPagesSerializerHelper
 ):
     child_sub_sectors = serializers.SerializerMethodField()
+    child_sub_articles = serializers.SerializerMethodField()
```
eadpearce: can this just be `articles` or `child_articles` pls

great_international/serializers.py
```diff
@@ -665,6 +666,13 @@ def get_child_sub_sectors(self, obj):
             MinimalPageSerializer
         )
 
+    def get_child_sub_articles(self, obj):
```
MichaelWalker: Do we also need to add a subscription for the cache busting? (or does this already work?)

sdonk: Forgot about that, good catch!

sdonk: Somehow it's already there https://github.com/uktrade/directory-cms/blob/develop/great_international/cache.py#L14 @MichaelWalker

PR #603, opened 2019-08-13
Tree based breadcrumbs serialize breadcrumbs label to be title if lab…

jessgilbert:
…el given

Not a breaking change as serializing `breadcrumbs_label` as `title` so still works with `cms_breadcrumbs`. 

 - [x] Changelog entry added.

### DIFF
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -10,6 +10,11 @@ class PageTitleAndUrlSerializer(serializers.Serializer):
     url = serializers.CharField()
 
 
+class PageBreadcrumbsAndUrlSerializer(serializers.Serializer):
+    title = serializers.CharField(source='breadcrumbs_label')
+    url = serializers.CharField()
+
+
 class BasePageSerializer(serializers.Serializer):
     id = serializers.IntegerField()
     seo_title = serializers.CharField()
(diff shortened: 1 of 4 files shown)


core/serializers.py
```diff
@@ -27,7 +32,15 @@ def get_tree_based_breadcrumbs(self, instance):
...
             page.specific for page in instance.specific.ancestors_in_app]
         breadcrumbs.append(instance)
-        return PageTitleAndUrlSerializer(breadcrumbs, many=True).data
+        serialized = []
+
+        for crumb in breadcrumbs:
+            if hasattr(crumb, 'breadcrumbs_label'):
+                serialized.append(PageBreadcrumbsAndUrlSerializer(crumb).data)
+            else:
+                serialized.append(PageTitleAndUrlSerializer(crumb).data)
```
richtier: can you add a test for this please :)

richtier: two serializers is a find approach. just FYI, we could also do:

```
class PageTitleAndUrlSerializer(serializers.Serializer):
    title = serializers.SerializerMethodField()
    url = serializers.CharField()

   def get_title(self, obj):
       if hasattr(obj, 'breadcrumbs_label'):
          return obj.breadcrumbs_label
       return obj.title
```

the advantages are:
  - the breadrumbs serialization logic lives in the breadcrumbs serializer. the caller does not need to know about the callee
 - reduced fragmentation: one way of doing breadcrumb serialization. therefore less things for devs to keep in mind.

jessgilbert: I tried this before and couldn't get it to work for some reason, have just tried it again. At that point would it be breadcrumbs labels with the country codes on?

jessgilbert: Ignore me had been doing it wrong 👍

PR #606, opened 2019-08-14
Added AboutUkRegionListingPage` and `AboutUkRegionPage

jessgilbert:

 - [x] Changelog entry added.


great_international/cache.py
```diff
@@ -236,6 +237,21 @@ class AboutUkLandingPageSubscriber(
...
+    subscriptions = [
+        great_international.AboutUkLandingPage
+    ]
+
+
+class AboutUkRegionPageSubscriber(AbstractDatabaseCacheSubscriber):
+    model = great_international.AboutUkRegionPage
+    subscriptions = [
+        great_international.AboutUkLandingPage,
+        great_international.AboutUkRegionListingPage
```
MichaelWalker: These all seem to be referencing parents. Is this so that we can do the breadcrumbs in the CMS?

If so that's a sad thing - means that any change to the top level model triggers the cache to be regenerated on ALL of its children. which might take ages (and normally not do much good)

(not suggesting that you do anything about this here - just a thing we should perhaps think about more carefully)

jessgilbert: Yes, cms breadcrumbs is the reason for it, but as we may not be pushing to use these breadcrumbs all the time yet maybe it's not worth doing this at the minute and using the other breadcrumbs

PR #617, opened 2019-08-22
Staging release

eadpearce:


great_international/panels/great_international.py
```diff
@@ -1185,6 +1186,7 @@ class AboutDitServicesPagePanels:
 class AboutUkLandingPagePanels:
 
     image_panels = [
+        FieldPanel('title'),
```
jessgilbert: This shouldn't be in image_panels, should be in content_panels

PR #618, opened 2019-08-23
Add migrations to remove markdown slug hyperlinks validator

eadpearce:

 - [x] Changelog entry added.


db_template.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 10.6 (Ubuntu 10.6-0ubuntu0.18.04.1)
--- Dumped by pg_dump version 10.6 (Ubuntu 10.6-0ubuntu0.18.04.1)
+-- Dumped from database version 9.6.14
```
MichaelWalker: Slightly interesting... 
Probably fine, (and less wrong than the current version so it shouldn't hold up the PR)

Does this imply that we are all using different versions of postgres locally?
If we are, then this is going to break things one day. Perhaps we should talk about it at the next dev team meeting.

Our README says that 9.5.x is the correct version to use for the CMS.

eadpearce: yeah agree. may be issues between v9/10 but for different minor versions of 9 it may not be a huge issue (i hope)

MichaelWalker: its possible that our dump uses a feature that's new to 9.6 and doesn't exist in 9.5.
but I agree that in practice our use of the DB is pretty basic and we are unlikely to be using cutting edge features. 😊

PR #621, opened 2019-08-23
Ci400 region map

jessgilbert:

 - [x] Changelog entry added.


great_international/models/great_international.py
```diff
@@ -1416,14 +1416,62 @@ class AboutUkLandingPage(panels.AboutUkLandingPagePanels, BaseInternationalPage)
     industries_section_cta_link = models.CharField(max_length=255, blank=True)
 
     regions_section_title = models.CharField(max_length=255, blank=True)
-    regions_section_content = MarkdownField(blank=True)
```
MichaelWalker: Just to check...

Deleting these is fine right, because there aren't any pages in production that use this template?

jessgilbert: Yep, not on staging or live, will it cause an error on dev if it has data for this field?

MichaelWalker: 👍 

Depends on the front end code, but probably!
The error will only be seen when loading that particular page though, so that's probably fine - lets just merge in the front end shortly after the cms

jessgilbert: Yeah the front-end doesn't depend on this or use this in the view or anything, tested it and it should be all good

great_international/serializers.py
```diff
@@ -1964,6 +2017,22 @@ class AboutUkRegionListingPageSerializer(BasePageSerializer):
...
+            data = [
+                AboutUkRegionsProxyDataWrapper(
+                    instance=queryset,
+                    region_title=region
+                )
+                for region in ['scotland', 'northern_ireland', 'north_england', 'wales', 'midlands', 'south_england']
+            ]
+            serializer = AboutUkRegionSerializer(data, many=True)
+            return serializer.data
+        return []
```
MichaelWalker: (not a big deal - don't feel like you have to change anything)

A minor code style point here is that you can reduce nesting and make it slightly easier to read by re-ordering this a little

```python
if not queryset:
    return []

data = ### some stuff
return serializer.data
```

it doesn't make a huge difference, but its slightly easier to read, as the effect of having no queryset is right next to the check, rather than 10 lines later.

great_international/serializers.py
```diff
@@ -1964,6 +2017,22 @@ class AboutUkRegionListingPageSerializer(BasePageSerializer):
...
+
+    def get_mapped_regions(self, instance):
+        queryset = AboutUkLandingPage.objects.live().public().first()
+        if queryset:
+            data = [
+                AboutUkRegionsProxyDataWrapper(
+                    instance=queryset,
+                    region_title=region
+                )
+                for region in ['scotland', 'northern_ireland', 'north_england', 'wales', 'midlands', 'south_england']
```
MichaelWalker: lets pull out these regions into a constant

tests/great_international/test_serializers.py
```diff
@@ -1118,3 +1118,124 @@ def test_invest_international_homepage_featured_industries(international_root_pa
...
+    )
+
+    assert len(serializer.data['mapped_regions']) == 6
+    for field in serializer.data['mapped_regions']:
+        assert 'region' in field
+        assert 'text' in field
+
+
+@pytest.mark.django_db
+def test_about_uk_landing_page_has_regions2(
```
MichaelWalker: slightly strange name for a test...

I don't think that this is really testing a different thing to the first one - I'd be tempted to combine the tests into a single one by adding the assert from the first test into this setup.

great_international/serializers.py
jessgilbert: 👍 Make sense will change as I prefer

great_international/serializers.py
MichaelWalker: In fact, can we pull out the whole `get_mapped_regions` method?

jessgilbert: Not quite I don't think, `get_mapped_regions` gets the regions form the about uk page that is live and includes that queryset, while `get_regions` is on the about uk page and just serializes it's own data

jessgilbert: Actually

jessgilbert:
```
regions = serializers.SerializerMethodField()

    def get_regions(self, instance):
        return get_mapped_regions(instance)
```

jessgilbert:
```
 mapped_regions = serializers.SerializerMethodField()

    def get_mapped_regions(self, instance):
        queryset = AboutUkLandingPage.objects.live().public().first()

        if not queryset:
            return []

        return get_mapped_regions(queryset)
```

jessgilbert:
```
def get_mapped_regions(instance):
    data = [
        AboutUkRegionsProxyDataWrapper(
            instance=instance,
            region_title=region
        )
        for region in REGIONS
    ]
    serializer = AboutUkRegionSerializer(data, many=True)
    return serializer.data
```

jessgilbert: Hows that?

great_international/serializers.py
```diff
@@ -1964,6 +2022,16 @@ class AboutUkRegionListingPageSerializer(BasePageSerializer):
...
 
+    mapped_regions = serializers.SerializerMethodField()
+
+    def get_mapped_regions(self, instance):
+        queryset = AboutUkLandingPage.objects.live().public().first()
+
+        if not queryset:
+            return []
+
+        return get_mapped_regions(queryset)
```
MichaelWalker: nice! :)

PR #631, opened 2019-09-02
CMS-1716 Remove obsolete code

sdonk:
To do:

 - [x] Changelog entry added.


core/cache.py
```diff
@@ -269,6 +267,11 @@ def build_slug_lookup_key(service_name, slug):
 
     @staticmethod
     def build_path_lookup_key(site_id, path):
+        path = path.strip('/')
+        if not path.endswith('/'):
```
MichaelWalker: we've just done the `strip('\')` so we know that this statement will always be true.

core/views.py
```diff
@@ -304,20 +289,28 @@ def dispatch(self, *args, **kwargs):
...
+
     def get_page(self):
         page_class = self.page_content_type.model_class()
-        try:
-            page = page_class.objects.get(
-                slug=self.request.POST.get('slug'),
-                service_name__iexact=self.kwargs['service_name']
-            )
-        except page_class.DoesNotExist:
+        # import pdb
```
sdonk: naughty @richtier

core/views.py
```diff
@@ -115,7 +117,7 @@ def check_parameter_validity(self):
         the endpoint an opportunity to raise exceptions due to invalid
         parameters values being supplied.
         """
-        self.get_object_id()
+        self.object_id
```
MichaelWalker: I'm probably missing some wagtail magic here....

but what's the point of caching this property? It looks like a simple getter, which doesn't do anything interesting and ought to return basically instantly.

core/models.py
```diff
@@ -236,7 +236,7 @@ def ancestors_in_app(self):
     @property
     def full_path(self):
         """Return the full path of a page, ignoring the root_page and
-        the app page. Used by the lookup-by-url view in prototype mode
+        the app page. Used by the lookup-by-path view in prototype mode
```
eadpearce: you can remove the `in prototype mode` part now

core/templates/core/upstream.html
```diff
@@ -5,20 +5,27 @@
     {% trans "Copy" as copy_str %}
     {% include "wagtailadmin/shared/header.html" with title=copy_str subtitle=page.get_admin_display_title icon="doc-empty-inverse" %}
     <div class="nice-padding">
-        <form id="form" path="{% url 'preload-add-page' service_name=service_name model_name=model_name parent_slug=parent_slug %}" method="POST" novalidate>
+        {% if not site %}
+            Cannot perform ation because the page must be attached to a site
```
eadpearce: "action"?

core/views.py
```diff
@@ -178,64 +180,45 @@ def check_parameter_validity(self):
...
 
-        self.object_id = object_id
         return object_id
 
 
 class PageLookupByPathAPIEndpoint(DetailViewEndpointBase):
 
-    def get_object_id(self):
+    @cached_property
+    def object_id(self):
```
MichaelWalker: We'd end up caching a 404 response here - is that safe to do?

ie, if I try it, get a 404, create the page I'm trying to find and try again...
will this remember the 404 and just return that cached response, or will it create a new one of these objects and be fine?

(also, I'm assuming that 'site_id' and 'path' can't ever change for a given instance of this object, so this cache is safe... is that right?)

core/views.py
```diff
@@ -304,20 +289,28 @@ def dispatch(self, *args, **kwargs):
...
     def get_page(self):
         page_class = self.page_content_type.model_class()
-        try:
-            page = page_class.objects.get(
-                slug=self.request.POST.get('slug'),
-                service_name__iexact=self.kwargs['service_name']
-            )
-        except page_class.DoesNotExist:
+        # import pdb
+        # pdb.set_trace()
```
MichaelWalker: left-over debugging!

core/views.py
sdonk: @richtier did this change. What's the reason behind it?

tests/core/test_views.py
```diff
@@ -86,15 +81,15 @@ def test_api_translations_are_loaded_when_available(
...
+    query_params = {'service_name': cms.GREAT_INTERNATIONAL}
     query_params.update(languge_query_params)
     response = client.get(url, query_params)
     assert response.status_code == 200
     assert response.json()['title'] == expected_title
 
 
 @pytest.mark.parametrize('language_code', (
-    'en-gb' 'de' 'ja', 'zh-hans', 'fr', 'es', 'pt', 'ar',
+    'en-gb', 'de', 'ja', 'zh-hans', 'fr', 'es', 'pt', 'ar',
```
MichaelWalker: did we decide that we couldn't just not have parameters in the end?

core/cache.py
sdonk: Good point. I'll change it

core/templates/core/upstream.html
sdonk: @richtier

PR #638, opened 2019-09-09
Ci 452 new international home page

jessgilbert:

 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -808,6 +910,144 @@ def get_featured_links(self, instance):
...
+            context=self.context
+        ).data
+        return serialized
+
+    def get_related_page_expand(self, instance):
+        serialized = []
+        if not instance.related_page_expand:
+            return serialized
+        serialized = RelatedInvestHomePageSerializer(instance.related_page_expand.specific).data
+        return serialized
```
MichaelWalker: not a big deal at all, but...

```python
if not instance.related_page_expand:
    return []

return RelatedInvestHomePageSerializer(instance.related_page_expand.specific).data
```

Does the same thing without the noise of that `serialized` variable that doesn't really do anything. 😊

PR #639, opened 2019-09-09
Hero image renditions

eadpearce:

 - [ ] Changelog entry added.

Look here for an example of how these will be used: https://directory-components-dev.london.cloudapps.digital/full-width-banners/

### DIFF
--- a/core/serializers.py
+++ b/core/serializers.py
@@ -1,4 +1,5 @@
 from rest_framework import serializers
+from wagtail.images.api import fields as wagtail_fields
 
 from core import fields
 from great_international.models.capital_invest import \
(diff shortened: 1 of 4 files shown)


MichaelWalker:
At the risk of just permanently talking about doing things via composition rather than inheritance.... (and please just tell me no if you think this is a bad idea)

Could we do this as a field searializer rather than an inherited class?

ie 
```python
class MyClass():
    hero_image = HeroImageField()
```

I think this would
- allow us to have more than 1 of these fields (eg for a case study and a hero etc)
- make it more inline with how other fields work
- allow us to return a 'hero-image-object', which would give us a nice extendable interface (eg `hero_image.large` etc, and a nice place to eg store meta-data like an 'alt-text' once and have it apply to all the image sizes.

core/serializers.py
```diff
@@ -15,6 +16,20 @@ class PageBreadcrumbsAndUrlSerializer(serializers.Serializer):
     url = serializers.CharField()
 
 
+class HeroThumbnailSerializer(serializers.Serializer):
+    hero_image_thumbnail = wagtail_fields.ImageRenditionField('fill-640x360', source='hero_image')
```
MichaelWalker: Could we just pull this into the main `HeroSerializer`?

PR #640, opened 2019-09-10
XOT-1033 Domestic Homepage

sdonk:
To do:

 - [ ] Changelog entry added.
 - [ ] (if adding new models) Add an updated version of db_template.sql


core/serializers.py
```diff
@@ -122,3 +128,54 @@ def get_same_sector_opportunity_pages_data_for(
...
         return related_sectors_with_opps
+
+
+# block serializers
+
+class StreamChildBaseSerializer(serializers.Serializer):
+
+    def to_representation(self, instance):
+        """
+        instance is wagtail.core.blocks.stream_block.StreamValue.StreamChild
```
MichaelWalker: in that case could we rename `instance` to `stream_child`?

core/serializers.py
```diff
@@ -122,3 +128,54 @@ def get_same_sector_opportunity_pages_data_for(
...
+        instance.value is either an instance of StructValue, if struct block, or an instance of the model
+        """
+        ret = OrderedDict()
+        fields = self._readable_fields
+
+        for field in fields:
```
MichaelWalker: if we refactored this to something like

```python
def get_field_representation():
    # all the code goes here

class StreamChildBaseSerializer(serializers.Serializer):
    def to_representation(self, stream_child):
        return { field.field_name: get_field_representation(stream_child, field) for field in self._readable_fields }
```

That would extract out a nice boring helper function that we could easily unit test.
Do you think that would be worth doing for this slightly fiddly bit of code?

sdonk: I literally copied and pasted this method from DRF, in order to change one line https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L513

We could, although that won't give us an ordered dictionary (we can pass a generator for that) and I'm sure there are better ways of doing this. We can explore those if we decide to go with the block/stream fields implementation. Refactoring now is a bit premature IMHO

core/serializers.py
```diff
@@ -122,3 +128,54 @@ def get_same_sector_opportunity_pages_data_for(
...
+# block serializers
+
+class StreamChildBaseSerializer(serializers.Serializer):
+
+    def to_representation(self, stream_child):
+        """
+        instance is wagtail.core.blocks.stream_block.StreamValue.StreamChild
+        instance.value is either an instance of StructValue, if struct block, or a single value
+        """
+        ret = OrderedDict()
```
richtier: fun fact: when we get to python 3.7 we can rely on insertion order :)
https://stackoverflow.com/a/39980548/904887

PR #642, opened 2019-09-10
Ci 452 new international home page

jessgilbert:


great_international/serializers.py
```diff
@@ -1024,27 +1023,35 @@ def get_all_sectors(self, instance):
...
         return serialized
 
     def get_related_page_buy(self, instance):
         serialized = []
         if not instance.related_page_buy:
             return serialized
-        serialized = RelatedTradeHomePageSerializer(instance.related_page_buy.specific).data
+        trade_page = RelatedTradeHomePageSerializer(instance.related_page_buy.specific).data
+        if 'image' in trade_page and 'title' in trade_page \
+                and trade_page['image'] and trade_page['title']:
```
richtier: 120 max line length now so this can be one line I think

PR #643, opened 2019-09-10
Added seo optimisation panels to more pages

jessgilbert:

 - [x] Changelog entry added.


great_international/panels/capital_invest.py
```diff
@@ -340,7 +343,8 @@ class CapitalInvestOpportunityListingPagePanels:
     content_panels = [
         FieldPanel('title'),
         FieldPanel('breadcrumbs_label'),
-        FieldPanel('search_results_title')
+        FieldPanel('search_results_title'),
+        SearchEngineOptimisationPanel()
```
MichaelWalker: We seem to sometimes put this in the settings_panels and sometimes in the content_panels.
It seems slightly odd that it isn't in the same place all of the time? Is there a reason for that?

jessgilbert: I can only see two places that it isn't in the content panel. Those two places may be in settings rather than content panels for a reason, but I don't know. I just stuck with the majority

eadpearce: translations is the reason 💀

jessgilbert: @eadpearce Do we want to keep it in the content panels then?

PR #650, opened 2019-09-13
Fix coverage

richtier:


.circleci/config.yml
```diff
@@ -35,18 +22,17 @@ jobs:
...
           command: |
             . venv/bin/activate
-            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term
+            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term --junitxml=test-reports/junit.xml
             codecov
       - store_test_results:
-          path: tests
-
+          path: test-reports
+      - store_artifacts:
```
richtier: allows us to optimize flaky and slow tests:
https://circleci.com/docs/2.0/collect-test-data/#pytest

.circleci/config.yml
```diff
@@ -1,24 +1,11 @@
...
 jobs:
   test:
-    <<: *machine_python3_postgres_redis
+    docker:
+      - image: circleci/python:3.6.6
+      - image: circleci/redis:3.2-alpine
     steps:
-      - *step_checkout_repo
-      - setup_remote_docker:
-          docker_layer_caching: true
```
richtier: by my reading of the docs, this does nothing https://circleci.com/docs/2.0/docker-layer-caching/#limitations

could be wrong though. let's see the effect on CI time

PR #653, opened 2019-09-16
Reinstate db template

eadpearce:

 - [x] Changelog entry added.


eadpearce:
migrations already fixed in another PR. making new branch

PR #671, opened 2019-09-18
Ci 502 update why choose uk page for ready to trade

jessgilbert:

 - [x] Changelog entry added.


great_international/panels/great_international.py
```diff
@@ -1581,6 +1581,8 @@ class AboutUkWhyChooseTheUkPagePanels:
         MultiFieldPanel(
             heading="Hero",
             children=[
+                HelpPanel('The hero title no longer shows, use this as alt text for the hero image which includes '
+                          'the text heading'),
```
MichaelWalker: This sounds like a little bit of a hack! 😊
Could you make sure there is a ticket to clean up after this?

jessgilbert: Yep ok

PR #678, opened 2019-09-23
XOT-1074 Add articlepage to the children of topiclandingpage

sdonk:

 - [x] Changelog entry added.


export_readiness/models.py
```diff
@@ -235,6 +235,7 @@ class TopicLandingPage(panels.TopicLandingPagePanels, BaseDomesticPage):
         'export_readiness.ArticleListingPage',
         'export_readiness.SuperregionPage',
         'export_readiness.CountryGuidePage'
+        'export_readiness.ArticlePage'
```
eadpearce: comma missing on line above

PR #680, opened 2019-09-24
Updated help panels

jessgilbert:

 - [x] Changelog entry added.


great_international/panels/great_international.py
```diff
@@ -1581,8 +1581,6 @@ class AboutUkWhyChooseTheUkPagePanels:
         MultiFieldPanel(
             heading="Hero",
             children=[
-                HelpPanel('The hero title no longer shows, use this as alt text for the hero image which includes '
-                          'the text heading'),
```
jessgilbert: This was the RTT change which never happened - hero is as normal

PR #683, opened 2019-09-25
Improve loading of db fixtures

eadpearce:

 - [x] Changelog entry added.


README.md
```diff
@@ -51,17 +51,20 @@ Secrets such as API keys and environment specific configurations are placed in `
...
 ### Setting up the local database
 
     $ make database
+
+`make database` drops then recreates the local database then loads `db_template.sql` so no migrations need to be run. Do this if you need a fresh database with no pre-existing pages or users.
+
     $ make load_fixtures
 
-`make database` drops the local database then loads `db_template.sql`. Then `make load_fixtures` loads fixtures from `fixtures/data.json`. This includes a dummy account with username "dev" and password "password".
+To add some dummy content to your local database run `make load_fixtures`. *This will overwrite your local database* with data from `db_fixtures.sql`. This includes a dummy account with username "dev" and password "password".
```
MichaelWalker: just `make fixtures` now?

makefile
```diff
@@ -60,7 +61,14 @@ database_template:
 		--dbname=cms_temporary_template
 	dropdb -U postgres cms_temporary_template
 
-load_fixtures:
-	ENV_FILES='secrets-do-not-commit,dev' ./manage.py loaddata fixtures/data.json
+fixtures:
+	psql -U postgres -h localhost -d directory_cms_debug -f db_fixtures.sql
+	ENV_FILES='secrets-do-not-commit,dev' \
+	export DATABASE_URL=postgres://postgres:postgres@localhost:5432/directory_cms_debug; \
```
MichaelWalker: is the database url not already in the env files?

eadpearce: I'd rather explicitly set it here because this command won't work if e.g. it's using the test env file

eadpearce: (since test env file uses sqlite and not postgres)

README.md
```diff
@@ -51,17 +51,20 @@ Secrets such as API keys and environment specific configurations are placed in `
 | make secrets                  | Create your secret env var file |
 | make worker                   | Run async cache celery worker |
 | make database                 | Updates the db template with any newly added migrations |
-| make database_template        | Drop, then set up fresh db from template |
+| make db_template              | Drop, then set up fresh db from template |
```
richtier: it's `database_template` https://github.com/uktrade/directory-cms/blob/develop/makefile#L51 :)

makefile
```diff
@@ -14,17 +14,15 @@ flake8:
 	--exclude=.venv,venv,node_modules,migrations \
 	--max-line-length=120
 
-ENV_FILES?='secrets-do-not-commit,dev'
 manage:
-	ENV_FILES=$(ENV_FILES) ./manage.py $(ARGUMENTS)
+	ENV_FILES='secrets-do-not-commit,dev' ./manage.py $(ARGUMENTS)
```
richtier: was this removed on purpose? prior to this change we could override the `ENV_VARS` used in when calling this e.g, `ENV_FILES=gleh make manage ...`

README.md
eadpearce: tab completion is faster when it's called "db_template". fight me

makefile
eadpearce: whenever i've run `make manage` this line `ENV_FILES?='secrets-do-not-commit,dev'` is ignored and the command uses `'test,dev'`. am i missing something?

README.md
richtier: no u

PR #684, opened 2019-09-25
Ci 518 add new how dit help section to int home page

jessgilbert:

 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -840,6 +840,24 @@ class InternationalHomePageSerializer(PageWithRelatedPagesSerializer, HeroSerial
...
+    related_how_dit_help_pages = serializers.SerializerMethodField()
+
+    def get_related_how_dit_help_pages(self, instance):
+        serialized = []
+        items = [
+            instance.related_how_dit_help_page_one,
+            instance.related_how_dit_help_page_two,
+            instance.related_how_dit_help_page_three
+        ]
+        for related_page in items:
```
eadpearce: you could do this as a one-liner:
`return [AboutDitServicesPageSerializer(related_page.specific).data for related_page in items if related_page]`

PR #690, opened 2019-10-01
Added new fields for new expand home page

jessgilbert:

 - [x] Changelog entry added.


great_international/panels/invest.py
```diff
@@ -24,6 +25,8 @@ class InvestInternationalHomePagePanels:
                 FieldPanel('breadcrumbs_label'),
                 FieldPanel('heading'),
                 FieldPanel('sub_heading'),
+                HelpPanel('Cta\'s require both text and a link to show '
```
MichaelWalker: This works fine, but a slightly neater trick is to just use different types of quotes

ie I'd do this instead.
```python
"Cta's"
```

PR #704, opened 2019-10-16
XOT-1161 change fields to optional on market pages

sdonk:

 - [ ] Changelog entry added.


export_readiness/models.py
```diff
@@ -286,7 +286,7 @@ def articles_count(self):
 
 
 class CountryGuidePage(panels.CountryGuidePagePanels, BaseDomesticPage):
-    """Make a cup of tea, this model is BIG!"""
+    """Make a cup of tea, this model is BIG! ☕"""
```
richtier: errant `=`

nice docstring tho :laughing:

sdonk: It's not an `=` it's a coffee cup emoji!
Screenshot 2019-10-16 at 14 55 25

PR #716, opened 2019-10-29
XOT-1177 upgrade wagtail 2.6

sdonk:

 - [x] Changelog entry added.


core/static/core/css/global.css
```diff
@@ -5,37 +5,6 @@
     color: #aaa;
 }
 
-/**
```
sdonk: https://github.com/uktrade/directory-cms/pull/361/commits/04daa462c371b83ed92fd55a179c30049d72cd22

PR #719, opened 2019-11-15
Xot 1185 remove market name prefix

jeanettemariemurphy:
…th 'the' alphabetically by country, without changing the display name

 - [x] Changelog entry added.

### DIFF
--- a/export_readiness/serializers.py
+++ b/export_readiness/serializers.py
@@ -504,6 +504,10 @@ class ChildCountryGuidePageSerializer(BasePageSerializer):
     hero_image_thumbnail = wagtail_fields.ImageRenditionField('fill-640x360', source='hero_image')
     heading = serializers.CharField()
     sub_heading = serializers.CharField()
+    sorted_title = serializers.SerializerMethodField()
+
+    def get_sorted_title(self, parent):
+        return parent.title.lower().replace('the ', '')
 
 
 class TopicLandingPageSerializer(BasePageSerializer, ChildPagesSerializerHelper, HeroSerializer):
--- a/export_readiness/serializers.py
+++ b/export_readiness/serializers.py
@@ -15,6 +15,8 @@ from .models import (
     ArticleListingPage, ArticlePage, CampaignPage, SuperregionPage, EUExitDomesticFormPage, CountryGuidePage
 )
 
+import re
+
 
 class RelatedArticlePageSerializer(BasePageSerializer):
     """Separate serializer for related article pages so we don't end up with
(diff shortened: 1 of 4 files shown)


export_readiness/serializers.py
```diff
@@ -504,6 +504,10 @@ class ChildCountryGuidePageSerializer(BasePageSerializer):
     hero_image_thumbnail = wagtail_fields.ImageRenditionField('fill-640x360', source='hero_image')
     heading = serializers.CharField()
     sub_heading = serializers.CharField()
+    sorted_title = serializers.SerializerMethodField()
+
+    def get_sorted_title(self, parent):
+        return parent.title.lower().replace('the ', '')
```
richtier: similar to https://en.wikipedia.org/wiki/Scunthorpe_problem

this will have unwanted side effects for words ending in `the` for example:

`The British Pathe videos => british pa videos`

This would affect the following words: https://www.wordfind.com/ends-with/the/

Also `lower()` would have unwanted effects for acronyms:

`The CIA worldbook => cia worldbook`

regex can solve both :)

```
return re.sub(r'^the', '', parent.title, flags=re.IGNORECASE)
```

jeanettemariemurphy: @richtier this change does not affect the display name, as this is still using the 'title' field.  'sorted_title' is just for sorting on country names, so The Netherlands will still display on the UI as The Netherlands, but is alphabetically sitting under N.
The lower case was just to prevent a sort deprioritorising if the cms returned The netherlands, for example
Using a regex would still be a good idea, but i wasn;t sure if it was necessary in this case?

eadpearce: off the top of my head I don't know if any country name ends in 'the' BUT if there are some like this maybe it'd be better to only look for it at the start of the string

jeanettemariemurphy: happy to change to the regex, but the point being that even if a country did end in 'the', it wouldn't affect the logic of what is trying to be achieved, ie the last 3 letters would be removed from the word , which would still place them correctly alphabetical order. And the UI display would not be affect at all.

eadpearce: yeah actually it's probably too much of an edge case to be an issue

tests/export_readiness/test_serializers.py
```diff
@@ -94,6 +94,29 @@ def test_country_guide_page_serializer(root_page, rf):
...
+        title='The Netherlands',
+        slug='country',
+        parent=markets_page)
+
+    serializer = ChildCountryGuidePageSerializer(
+        instance=country_guide,
+        context={'request': rf.get('/')}
+    )
+
+    sorted = serializer.data['sorted_title']
```
eadpearce: `sorted` is a builtin python function. can you rename to something like `sorted_title`? https://docs.python.org/3.5/library/functions.html#sorted

PR #723, opened 2019-12-03
XOT-1232 Market pages filtering API

sdonk:
 - [x] Changelog entry added.


conf/urls.py
```diff
@@ -59,6 +59,20 @@
         ),
         name='industry-tags-list'
     ),
+    url(
+        r'^pages/lookup-countries/$',
+        api_router.wrap_view(
+            export_readiness.views.CountryPageListAPIView.as_view()
```
richtier: 120 chars should allow this to be one lline:
```
         api_router.wrap_view(export_readiness.views.CountryPageListAPIView.as_view())
```

export_readiness/serializers.py
```diff
@@ -576,12 +577,12 @@ def get_countries(self, object):
...
 
 
-class TagSerializer(serializers.Serializer):
+class IDNameSerializer(serializers.Serializer):
     id = serializers.IntegerField()
     name = serializers.CharField()
 
 
-class IndustryTagSerializer(TagSerializer):
+class IndustryTagSerializer(IDNameSerializer):
```
richtier: I guess inheriting IDNameSerializer to be DRY.

seems a bit overkill. There's DRY and then there's SAND: tiny bit of DRY things.

I don't like sand. It's coarse and rough and irritating and it gets everywhere

PR #738, opened 2019-12-11
Production release

sdonk:
 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -458,7 +419,9 @@
     WAGTAIL_PASSWORD_MANAGEMENT_ENABLED = False
     WAGTAIL_PASSWORD_RESET_ENABLED = False
     WAGTAILUSERS_PASSWORD_ENABLED = False
-
+    MIGRATE_EMAIL_USER_ON_LOGIN = env.bool('MIGRATE_EMAIL_USER_ON_LOGIN', False)
+else:
+    LOGIN_URL = '/admin/login/'
```
sdonk: I think we need this even when SSO is on @richtier

richtier: when staff sso is on it's set to something else: https://github.com/uktrade/directory-cms/pull/738/files#diff-e398a065684e871bec35f76ea80f20a7R407

PR #742, opened 2019-12-17
Update pillow to fix vulnerability

eadpearce:

 - [x] Changelog entry added.
 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [x] (if updating requirements) Requirements have been compiled.


requirements.txt
```diff
@@ -54,7 +54,7 @@ monotonic==1.5            # via notifications-python-client
 more-itertools==8.0.0     # via zipp
 notifications-python-client==5.3.0
 oauthlib==3.1.0           # via requests-oauthlib
-pillow==6.2.1             # via wagtail
+pillow==6.2.1
```
sdonk: 🤔 it's the same version!

eadpearce: the vulnerability alert was specifically for requirements_test... i wonder why there's a version mismatch between the two

eadpearce: could be someone only compiling for requirements and not requirements_test

PR #810, opened 2020-03-10
Cache /api/pages/<pk>/

richtier:

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -80,42 +80,6 @@ class APIEndpointBase(PagesAdminAPIEndpoint):
         self.handle_activate_language(instance)
         return instance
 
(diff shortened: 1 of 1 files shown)


core/views.py
```diff
@@ -80,14 +80,57 @@ def get_object(self):
...
+        self.check_parameter_validity()
+
+        if helpers.is_draft_requested(request):
+            return super().detail_view(request, pk=None)
+
+        # Return a cached response if one is available
+        cached_data = cache.PageCache.get(
+            page_id=self.object_id,
+            lang=translation.get_language(),
+            region=request.GET.get('region'),
```
sdonk: Did we not kill region? 🤔

PR #867, opened 2020-12-04
GBAU-887: Ensure MarketingArticlePages are also sent to ActivityStream

stevejalim:
This changeset allows additional Campaign article pages to appear in search results.

Before this changeset, `ArticlePage`s with the type set to `Campaign` were already being sent to ActivityStream (and therefore appearing in search results via domestic-ui) but VERY similar pages using the alternative `MarketingArticlePage` model were not.

This is addressed by ensuring that page class is also queried for and serialized for sending to AS.

Note that, in common with the current pattern, all pages are being send with the same type of '`Article`', so no change was needed in terms of how we query for these pages. It also means that the pages are labelled as 'Article' not 'Marketing' in search results -- I've checked with a PO who understands BAU and he's happy with this.

In addition to the unit test, I've also checked this manually by connecting my local build to Dev's ActivityStream

 - [X] Changelog entry added.

**PLEASE DO NOT MERGE UNTIL MONDAY**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Pre-release
 
 ### Implemented enhancements
+- GBAU-887 - Send MarketingArticlePages to ActivityStream so they appear in search results
 - GAA-31 - Upgrading wagtail to 2.10
 - no-ticket - migrate to codecov remove coveralls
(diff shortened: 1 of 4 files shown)


depsiatwal (approved):
LGTM

PR #890, opened 2021-08-13
GP2--3153-3154--new-opportunity-pages

stevejalim:
This changeset is a squashed set of commits focused on adding two new page types to directory-cms, for the atlas project.

* A new InvestmentOpportunityPage (which replaces the concept of Capital Investment Opportunties and High Potential Opportunties being held in separate models, instead handling them both in the same page model)
* Its parent InvestmentOpportunityListPage.

These are similar to the detail and listing pages in the capital_invest module, but not identical. They do take much of the relationship modelling from them, so we can reuse as much code in great-international-ui as possible - we will try to repurpose much of the view and template code for capinvest for the atlas pages.

Other points to note:
* The new detail page uses a StreamField.
* Rich text is still done with markdown, which is rendered to HTML before it goes over the wire.
* Wagtail-modeltranslation has been configured for the new pages, and it appears to play nicely with streamfields. Note that for the MVP atlas, we won't have translations being used.
* The changeset also includes Serializers (and nested Serializers) for the new model, along with SOME tests, covering critical behaviour. Coverage was not at 100% before this work started, but we should aim to get it up later on.
* Migrations (10+) have been merged/squashed/recreated as a single migration, because running each migration in d-cms takes time
* Coverage.py is run by default with pytest now

 - [X] Changelog entry added.


great_international/blocks/investment_atlas.py
```diff
@@ -0,0 +1,106 @@
...
+        retval_dict = {}
+
+        img = value.get("image")
+
+        if img:
+            _rendition_spec = context.get('rendition_spec', DEFAULT_IMAGE_RENDITION_SPEC)
+            retval_dict.update(
+                {
+                    'image': img.get_rendition(_rendition_spec).url,
+                    'image_alt': value.get('image_alt', ''),
```
webbyfox, line 40: `image_alt` is required and also .get will give null string if empty
```suggestion
                    'image_alt': value.get('image_alt'),
```

tests/great_international/factories.py
```diff
@@ -809,3 +809,70 @@ class Meta:
...
+
+    planning_status = factory.SubFactory(PlanningStatusFactory)
+    investment_type = factory.SubFactory(InvestmentTypeFactory)
+    time_to_investment_decision = models.investment_atlas.TIME_TO_INVESTMENT_DECISION_0M_6M
+
+    # Streamfields - set them in the instance
+    # featured_images
+    # main_content
+
+    # related_regions - set them on the instance from the factory
```
webbyfox, line 855: Are the above comments still valid? I think code has moved somewhere but not comment

webbyfox (approved):
Overall looks good. Just minor comments but not blocker to approve this

tests/great_international/factories.py
stevejalim: Yeah they are still valid - featured_images and main_content are streamfields, and related_regions is a ParentalManyToManyField on InvestmentOpportunityPage

PR #892, opened 2021-08-18
GP2 3157: Add Investment Atlas homepage

stevejalim:
This page goes as a child of the international homepage, and has (for now, just) the investment opportunities listing page as its child.

Changes made:

* add new model
* add new streamfield blocks
* add new translation config
* add new wagtail panel block
* add new Migration
* Parent/sub-page permission tweaks
* Added Serializer so the page will be sent to the redis cache

Manually tested

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #893, opened 2021-08-20
NOTICKET: Fix failing PutObject call from Wagtail

stevejalim:
This is the result of a pairing discussion to work out why Wagtail couldn't generate thumbnails, because it was being denied (403) for PutObject

Changes kept to a minimum

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #895, opened 2021-08-23
Amended full url for Atlas homepage and opportunities listing page

webbyfox:
This changeset amends URL fields in the serializer. These homepage doesn't needs to have content in url and have their own standalone page.


froddd (approved):
LGTM

PR #896, opened 2021-08-30
GP2-3330: Extend PlanningStatus snippet model field to support a verbose description

stevejalim:
This changeset expands the representation of a PlanningStatus snippet so that we store a verbose name (for display in the main opportunity page) as well as the short version (for use in the sidebar filter)

New representation of the data

```
{
     ...
     "planning_status": {
          "name": "Planning Status example", 
          "verbose_description": "Planning Status example verbose desc here"
     },
     ...
}
```

 - [X] Changelog entry added.


depsiatwal (commented):
LGTM appears more tests are required though

depsiatwal (approved):

PR #897, opened 2021-08-31
GP2-3330: Extend InvestmentOpportunityForListPageSerializer to guarantee representation of PlanningStatus snippet

stevejalim:
By default, the serializers.CharField was using the string method of the FKed PlanningStatus record to get its name. This worked, but would break things if that __str__() method was changed.

This changeset makes things stable by explicitly stating what to use for the planning-status name when listing opportunities.

 - [X] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Pre-release
 ### Implemented enhancements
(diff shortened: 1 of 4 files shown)


depsiatwal (approved):
LGTM

PR #898, opened 2021-09-01
GP2-3306: Support reusable snippets in InvestmentOpportunityPage's main content streamfield

stevejalim:
This changeset adds a new snippet (ReusableContentSection) which contains a StreamField, allowing flexible content in that snippet.

An instance of a ReusableContentSection snippet can then be included in the InvestmentOpportunityPage.main_content StreamField.

This is so that an editor can define certain blocks of contnet in one place and then re-use them across multiple InvestmentOpportunityPages without having to re-key it all. They will, however, have to specifically select the relevant snippet to include in the InvestmentOpportunityPage's main_content streamfield

The `ReusableContentSection.content` StreamField has the same field spec as the main `InvestmentOpportunityPage.main_content` StreamField)

 - [X] Changelog entry added.


depsiatwal (approved):
LGTM

PR #900, opened 2021-09-03
Revert "GP2-3158: Refactor International Homepage for new content"

stevejalim:
Reverts uktrade/directory-cms#899 because deployment timed out, so we need to find a different way forward


depsiatwal (approved):
LGTM

PR #901, opened 2021-09-03
GP2-3158: [Revisited] Update international homepage CMS page model and serializer

stevejalim:
This changeset radically refactors/cuts down the CMS page serializer for the InternationalHomePage, ready for the new design.

It also adds a streamfield to InternationalHomePage, using the new InternationalHomepagePanelBlock.

IMPORTANT: There are a large number of fields which are no longer needed, including some which appear to be from an the previous incarnation of the homepage, but these could not be dropped because it created a migration that too ~22 mins to run and times out on the PaaS. As such, they are marked as redundant but they remain in the DB.

Note that until great-international-ui is updated to match the new data structure, this will effectively break great-international's homepage, so coordination of releases must be done.

How to test:

* Back up your local directory-cms database first
* Locally, pull the branch and run the migration
* Edit the Great International homepage in the CMS - confirm you can set a hero title and also a streamfield of 'panels'
* You can publish the page, but if you do so, note that this will break the page on the great-international-ui service, which has yet to be updated to suit the new data structures

 - [X] Changelog entry added.


depsiatwal (approved):
LGTM

PR #903, opened 2021-09-06
GP2-3287:  Add 'related opportunities' to the data available for a Region page

stevejalim:
This changeset expands (but does not otherwise amend) the API representation of a Region page, so that it includes up to three serializes Investment Opportunities in its response. (ie, the Opps are Atlas ones, not the old CapInvest type)

To be selected, an InvestmentOpportunity must have the current region as one of their related regions. They are then sorted by priority_weighting, if set, and then use their PK as a secondary sorting key. This means that highly weighted ones appear first, then any ties are broken by using the most recently created (ie highest PK) opp first.

In total up to three opps are included, but 0..3 is possible.

 - [X] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Pre-release
 ### Implemented enhancements
+- GP2-3287: Add Related Opportunities to the region-page's serializer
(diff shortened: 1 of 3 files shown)


webbyfox (approved):
LGTM

PR #905, opened 2021-09-07
Fixed InvestmentOpportunityForListPageSerializer when there is no pla…

webbyfox:
…nning_status

 - [x] Changelog entry added.


stevejalim (approved):
Thanks!

PR #908, opened 2021-09-08
GP2-3215: Add replacement Sector and SubSector pages

stevejalim:
This is a large changeset, but unavoidable. The original scope was to extend/refactor the existing `InternationalSectorPage` class to have new fields in line with the new design, but - as we discovered with InternationalHomePage - dropping translated fields creates huge migrations that take too long to run.

Instead, this changeset adds two new models that effectively replace `InternationalSectorPage` and `InternationalSubSectorPage` and then wires them in to relevant pages for Atlas. 

These new pages are called `InternationalInvestmentSectorPage` and `InternationalInvestmentSubSectorPage`.

Note that old pages (eg CapInvest pages) that are due for deprecation but which reference the old sector pages have NOT been updated. Pages we will continue to use - eg `InternationalTopicLandingPage` - HAVE been updated to use the new models.

Another thing to note is that `InvestmentOpportunityPage` was using a `RelatedSector` and `RelatedSubSector` Django models which in turn keyed to the OLD Sector pages. In order to switch the FKs from those models to point to the new models, I had to delete some data for `RelatedSector` and `RelatedSubSector` first. This is fine, because this data would only exist specifically for Atlas, and the content team know they will need to re-link InvestmentOpportunityPages to Sectors and SubSectors. When we've deployed this code to dev, we can double-check no pages have accidentally been deleted - eg page ID 787 exists now and should/will exist after this is deployed

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #909, opened 2021-09-09
NOTICKET: Atlas cms snags

stevejalim:
* remove word limit on intro field for opportunity
* add help_text for streamfields to show options for block_slug special values


great_international/models/investment_atlas.py
```diff
@@ -307,10 +314,9 @@ class InvestmentOpportunityPage(
     )
 
     introduction = MarkdownField(
-        max_length=300,
         blank=False,
         help_text=(
-            'A single paragraph of 300 characters max including spaces to introduce the opportunity '
+            'A single paragraph max including spaces to introduce the opportunity '
```
froddd, line 319: Some extra content still leftover here, I guess this should read `A single paragraph to introduce the opportunity `

froddd (approved):
Slight tweak needed to one help text, good to go once fixed

PR #910, opened 2021-09-09
GP2-3216: Add hero image for each region to AboutUKRegionSerizalizer

stevejalim:

 - [X] Changelog entry added.


agiamas (approved):
lgtm

PR #915, opened 2021-09-10
Added WhyInvestIntheUK model

webbyfox:
This changeset adds a new WhyInvestIntheUK page model. Added relevant serializer for FE and translation fields. Design is not finalised yet but we have rough copy to work on.

 - [x] Change has a Jira ticket that has the correct status.
 - [x] Changelog entry added.


great_international/models/great_international.py
```diff
@@ -2077,3 +2069,51 @@ class AboutUkWhyChooseTheUkPage(panels.AboutUkWhyChooseTheUkPagePanels, BaseInte
...
+        related_name='+',
+        blank=True
+    )
+
+    investment_opps_title = models.CharField(
+        max_length=255,
+        verbose_name="Investment opportunities title", blank=True
+    )
+    investment_opps_intro = models.CharField(
+        max_length=255,
```
stevejalim, line 2116: I'd say this is going to be too short - might be worth making it 1000 now (and this one should probably be a TextField?)

great_international/models/great_international.py
```diff
@@ -2077,3 +2069,51 @@ class AboutUkWhyChooseTheUkPage(panels.AboutUkWhyChooseTheUkPagePanels, BaseInte
...
+    hero_title = models.CharField(max_length=255)
+    hero_image = models.ForeignKey(
+        'wagtailimages.Image',
+        null=True,
+        on_delete=models.SET_NULL,
+        help_text='Main page hero image, above the title',
+        related_name='+'
+    )
+
+    featured_description = models.TextField(max_length=255, blank=True)
```
stevejalim, line 2089: I'd say this is going to be too short - might be worth making it 1000 now (and editing the migraiton to avoid having to make a second one)

great_international/models/great_international.py
```diff
@@ -2077,3 +2069,51 @@ class AboutUkWhyChooseTheUkPage(panels.AboutUkWhyChooseTheUkPagePanels, BaseInte
...
+        'wagtailimages.Image',
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+        blank=True
+    )
+    region_summary_section_intro = models.TextField(max_length=255, blank=True)
+
+    uk_sector_section_title = models.CharField(max_length=255, blank=True)
+    uk_sector_section_intro = models.TextField(max_length=255, blank=True)
```
stevejalim, line 2102: I'd say this is going to be too short - might be worth making it 1000 now (and editing the migraiton to avoid having to make a second one)

great_international/serializers.py
```diff
@@ -2663,3 +2663,23 @@ class InternationalInvestmentSubSectorPageSerializer(
     BasePageSerializer,
 ):
     heading = serializers.CharField()
+
+
+class WhyInvestInTheUKPageSerializer(
+    BasePageSerializer,
+    HeroSerializer,
+):
+    IMAGE_RENDITION_SPEC = "fill-500x500"
```
stevejalim, line 2672: Maybe make this a bit bigger? like 900w?

stevejalim (changes requested):
Looks great, but worth making some text char limits a bit broader sooner rather than later?

webbyfox:
Amended @stevejalim

stevejalim (approved):

PR #920, opened 2021-09-13
NOTICKET: Update downpage content block on Sector page

stevejalim:
This changeset gets rid of the separate text OR image blocks for the downpage content on the Sector page and instead supports a new text+optional image block

 - [X] Changelog entry added.


webbyfox (approved):
Lgtm

PR #922, opened 2021-09-14
Gp2 3374 refactor whyinvestintheuk page

webbyfox:
This changeset refactor WhyinvestintheUK page.

 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -2672,15 +2662,13 @@ class WhyInvestInTheUKPageSerializer(
 ):
     IMAGE_RENDITION_SPEC = "fill-960x540"
 
+    featured_summary = serializers.CharField()
     featured_description = serializers.CharField()
     hero_title = serializers.CharField()
     hero_image = wagtail_fields.ImageRenditionField(
```
stevejalim, line 2668: Because we're using the HeroSerializer here, we don't need the hero_image to be explicitly specced here

stevejalim (approved):
Minor comment but LGTM

great_international/serializers.py
webbyfox: Amended

PR #923, opened 2021-09-14
GP2-3400: add new general-use content page for Atlas content

stevejalim:
This changeset adds a general-purpose content page using a streamfield, which should support the 'How We Can Help' page for Atlas, as well as its child pages (hopefully)

There will be a corresponding change to GIUI to use this page, too

 - [X] Changelog entry added.


webbyfox (approved):
LGTM - just need to update migration file to avoid conflict

PR #927, opened 2021-09-16
GP2-3400: Update block spec for general Atlas content page to allow optional CTAs

stevejalim:
This change means the general atlas content page's stream field has the following block options:

* content block, containing:
   * Markdown/text
   * Image - (optional)
   * CTA link and label - (optional) 
* columns block, containing 1...n markdown columns 

This changeset includes making the ButtonBlock's label field totally optional, which allows it to be used as an overall optional sub-block

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #932, opened 2021-09-17
Gp2 3301 extend regions page for three pages in footer

webbyfox:
This changeset adds related page selection panel for the regions page
<details>
  <summary>Screenshots</summary>

image
</details>

 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -2012,7 +2037,7 @@ def get_regions(self, instance):
         return get_mapped_regions(instance)
 
 
-class AboutUkRegionListingPageSerializer(BasePageSerializer, HeroSerializer):
+class AboutUkRegionListingPageSerializer(PageWithRelatedPagesSerializer, HeroSerializer):
```
stevejalim, line 2040: As long as this change hasn't broken the Regions listing page, this is great

stevejalim (approved):
LGTM - nice one. Minor question/paranoia moment in a comment.

great_international/serializers.py
webbyfox: `PageWithRelatedPagesSerializer` inherits from `BasePageSerializer` so its fine

PR #938, opened 2021-09-21
GP2-3407: Rename HPO contact form as FDI form and relocate

stevejalim:
This changeset handles the backend refactoring for moving the CMS-driven High Potential Opportunities contact form into Investment Atlas, renaming it to Foreign Direct Investment instead of HPO.

Changes:
* all allow the form page to be a child of the Atlas landing page, but retain the success page as a child of the form page
* Move and rename all related logic from the `invest` app into the `investment_atlas` app, to make dead-code removal easier in the future
* add a schema migration that renames the relevant DB tables, in line with the refactor above
* update serizalizer to filter which opportunities to show based on whether or not their investment-type's label is Foreign Direct Investment. This value is controlled by an env-overrideable setting
* update serializer mapping
* update test factories
* update CMS editing/panel config
* update translation config (though note that these forms had no translation enabled for them - it was explicitly turned off for them)

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #942, opened 2021-09-23
NOTICKET: Increase length of case study text field on region page 

stevejalim:
Requested by content team - increasing from 255 to 500 chars

 - [X] Changelog entry added.


webbyfox (approved):
LGTM

PR #944, opened 2021-09-23
NOTICKET: make Region case study serializer match model char limit

stevejalim:

 - [X] Changelog entry already added.


webbyfox (approved):
LGTM

PR #946, opened 2021-09-27
Added Streamfield for locations for InvestmentOpportunityPage

webbyfox:
This changeset adds location to regions_with_locations as streamfield and links regions and map coordinates - Amended necessory serializer

 - [x] Changelog entry added.


great_international/serializers.py
```diff
@@ -2414,14 +2424,23 @@ class InvestmentOpportunityPageSerializer(BasePageSerializer, HeroSerializer):
...
-        many=True,
-        read_only=True,
-    )
     related_sectors = serializers.SerializerMethodField()
     sub_sectors = serializers.SerializerMethodField()
     related_opportunities = serializers.SerializerMethodField()
+    related_regions = serializers.SerializerMethodField()
+
+    def _get_regions(self, instance):
+        return set([item.value['region'] for item in instance.regions_with_locations])
```
stevejalim, line 2433: Because 'regions' is not a mandatory field in the structblock, it's possible someone will save bad data and make this blow up. Not an approval blocker, but worth a tweak - this might work:
```suggestion
        return set([item.value['region'] for item in instance.regions_with_locations if item.value['region']])
```

stevejalim: I guess the other helpers that do this will also benefit from the fix, too

great_international/serializers.py
```diff
@@ -2550,11 +2574,15 @@ class InvestmentOpportunityForListPageSerializer(BasePageSerializer):
...
     investment_type = serializers.CharField(max_length=255)
     time_to_investment_decision = serializers.SerializerMethodField()
+    locations_with_regions = StreamFieldSerializer(source="regions_with_locations")
 
     related_regions = serializers.SerializerMethodField()
     related_sectors = serializers.SerializerMethodField()
     sub_sectors = serializers.SerializerMethodField()
 
+    def _get_regions(self, instance):
+        return set([item.value['region'] for item in instance.regions_with_locations if item.value['region'].live])
```
stevejalim, line 2584:
```suggestion
        return set([item.value['region'] for item in instance.regions_with_locations if item.value['region'] and item.value['region'].live])
```

great_international/serializers.py
```diff
@@ -2564,7 +2592,16 @@ def get_time_to_investment_decision(self, instance):
...
+
+        related_regions = set()
+
+        for opp in InvestmentOpportunityPage.objects.live().public().order_by(
+            '-priority_weighting', '-pk'
+        ):
+            # We only want to add regions for THIS opportunity
+            if opp.pk == instance.pk:
+                related_regions.update(self._get_regions(opp))
+
```
stevejalim, line 2604: Looking at this again, can it not be replaced with simply:

```
related_regions = self._get_regions(instance)

```

stevejalim (approved):
Approved, but there are a couple of bits we could improve I think

stevejalim (approved):
Awesome, nice

PR #982, opened 2021-10-20
For Opp page, get related opps via sectors

webbyfox:

 - [ ] Changelog entry added.

### DIFF
--- a/great_international/serializers.py
+++ b/great_international/serializers.py
@@ -2501,22 +2501,21 @@ class InvestmentOpportunityPageSerializer(BasePageSerializer, HeroSerializer):
         return sub_sectors_list
 
     def get_related_opportunities(self, instance):
-        # Related opportunities are defined as ones in the same REGION, whereas previously (with the
-        # CapitalInvestOpportunityPageSerializer) we were showing based on SECTOR.
+        # Related opportunities are defined as based on SECTOR.
 
-        related_regions_set = self._get_regions(instance)
+        related_sectors_set = set(sector.related_sector_id for sector in instance.related_sectors.all())
 
-        if not related_regions_set:
+        if not related_sectors_set:
             return []
 
         related_opps = []
-
         for opp in InvestmentOpportunityPage.objects.live().public().exclude(
                 id=instance.id
         ).order_by(
             '-priority_weighting'
         ).distinct():
-            if self._get_regions(opp).intersection(related_regions_set):
+            opp_sectors_set = set(sector.related_sector_id for sector in opp.related_sectors.all())
+            if opp_sectors_set.intersection(related_sectors_set):
                 related_opps.append(opp)
 
         if not related_opps:
--- a/tests/great_international/test_serializers.py
+++ b/tests/great_international/test_serializers.py
@@ -1554,8 +1554,8 @@ def test_atlas_opportunity_page_null_case_related_sector__alt(rf, international_
 
 @pytest.mark.django_db
 def test_atlas_opportunity_list_page_related_opportunities(rf, international_root_page):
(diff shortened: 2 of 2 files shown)


great_international/serializers.py
```diff
@@ -2504,19 +2504,19 @@ def get_related_opportunities(self, instance):
         # Related opportunities are defined as ones in the same REGION, whereas previously (with the
         # CapitalInvestOpportunityPageSerializer) we were showing based on SECTOR.
```
froddd, line 2505: This comment can probably be removed, or at least updated!

tests/great_international/test_serializers.py
```diff
@@ -1848,19 +1873,18 @@ def test_atlas_opportunity_list_page_related_opportunities(rf, international_roo
...
     ]
 
     unexpected_slugs = [
         x.slug for x in [
             # priority doesn't matter here, because we're testing for absence
             opportunity_1,
-            opportunity_5,
-            opportunity_6,
+            opportunity_3,
+            opportunity_4,
```
froddd, line 1887: Should this list also include `opportunity_9` and `opportunity_10`?

tests/great_international/test_serializers.py
```diff
@@ -1624,26 +1622,25 @@ def test_atlas_opportunity_list_page_related_opportunities(rf, international_roo
...
+    # 1. Sector A, Subsector A1 -> Thing we'll match against
+    # 2. Sector A, no subsector -> valid match
+    # 3. Sector B, Subsector B2 -> valid match
+    # 4. Sector B, Subsector B3 -> valid match
+    # 5. Sector A, -> should not be shown
+    # 6. Sector A, Subsector A1 -> should not be shown
+    # 7. Sector C -> should be shown
+    # 8. Sector C -> should  be shown
+    # 9. Sector C -> valid match
+    # 10 with no sector -> should not be shown
```
froddd, line 1635: Let's use `related` and `not related` instead of `valid match`/`should not be shown`/`should be shown`

tests/great_international/test_serializers.py
```diff
@@ -1875,48 +1899,54 @@ def test_atlas_opportunity_list_page_related_opportunities(rf, international_roo
...
```
froddd, line 1914: Same as above -- this is a bit unclear. Could we comment with the expected number of related opps as this is what we're testing?

tests/great_international/test_serializers.py
froddd: Also I don't think the expected results here are correct -- I'd expect to see:

```suggestion
    # 2. Sector A, no subsector -> related
    # 3. Sector B, Subsector B2 -> not related
    # 4. Sector B, Subsector B3 -> not related
    # 5. Sector A, -> related
    # 6. Sector A, Subsector A1 -> related
    # 7. Sector C -> not related
    # 8. Sector C -> not related
    # 9. Sector C -> not related
    # 10 with no sector -> not related
```

tests/great_international/test_serializers.py
froddd:
```suggestion
    # Also show the number of related opps varies according to mutual sectors
    # 1. Sector A, Subsector A1 -> Thing we'll match against
    # 2. Sector A, no subsector -> 3 related opps
    # 3. Sector B, Subsector B2 -> 1 related opp
    # 4. Sector B, Subsector B3 -> 1 related opp
    # 5. Sector A, -> 3 related opps
    # 6. Sector A, Subsector A1 -> 3 related opps
    # 7. Sector C -> 2 related opps
    # 8. Sector C -> 2 related opps
    # 9. Sector C -> 2 related opps
    # 10 with no sector -> 0 related opps
```

PR #992, opened 2021-11-22
Removed CountryGuidePage

webbyfox:

 - [ ] Changelog entry added.


activitystream/serializers.py
```diff
@@ -58,5 +34,5 @@ def to_representation(self, obj):
             return ArticlePageSerializer(obj).data
         elif isinstance(obj, MarketingArticlePage):
             return MarketingArticlePageSerializer(obj).data
-        else:  # CountryGuidePage
-            return CountryGuidePageSerializer(obj).data
+        else:
+            return {}
```
depsiatwal, line 38: looks like tests may not be covering this

depsiatwal (commented):
just small comment regarding test

activitystream/serializers.py
webbyfox: I have added for safe side but I think its not needed

webbyfox: Removed @depsiatwal

depsiatwal (approved):
LGTM

PR #1003, opened 2022-01-11
GLS-45 Cleanup InternationalHomePage model from legacy fields

froddd:
Finish removing deprecated fields from `InternationalHomePage` model.

 - [x] Changelog entry added.


webbyfox (approved):
This might be treaky migration

PR #1026, opened 2022-07-14
Update select2 version from 4.0.6-rc.0 to 4.0.6

tomstevens9:

 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.


webbyfox (approved):
@tomstevens9  I will deploy this branch and test on dev before merging it

tomstevens9:
As far as I can tell the only place where we are using `select2` is on the access request page.

https://github.com/uktrade/directory-cms/blob/develop/users/forms.py#L117

I have added screenshots below to evidence that this select box is still working correctly.

Screenshot 2022-07-15 at 16 35 37
Screenshot 2022-07-15 at 16 36 18
Screenshot 2022-07-15 at 16 35 55

PR #1047, opened 2022-11-16
Add intro_video field to investment opportunity page

mforner13:
This PR adds a new `intro_video` field to the investment opportunity page. This will be displayed in place of an `intro_image` in the great-international-ui frontend when it is populated.

**Screenshot from wagtail**
Screenshot 2022-11-16 at 09 23 22

 - [x] Changelog entry added.


bsabhilash:
Looks really good

PR #1070, opened 2023-02-28
KLS-247: add hero_image and hero_video to article page for freeports work

LuptonM:

 - [ ] Changelog entry added.


mforner13 (approved):
LGTM 👍

great_international/migrations/0159_auto_20230228_1653.py
```diff
@@ -0,0 +1,151 @@
+# Generated by Django 3.2.18 on 2023-02-28 16:53
```
mforner13, line 1: Sorry missed this originally - as a rule we change the migration file name to describe the change, rather than leaving it as the auto generated name )(but keep the code at the beginning).

E.g. `0159_add_hero_image_video_article_page.py`

mforner13 (changes requested):

great_international/migrations/0159_auto_20230228_1653.py
LuptonM: now renamed

mforner13 (approved):

PR #1071, opened 2023-03-08
Added freeport data into article page serialiser

LuptonM:

 - [ ] Changelog entry added.


great_international/serializers.py
```diff
@@ -312,6 +312,22 @@ class InternationalArticlePageSerializer(BasePageSerializer):
 
     tags = core_fields.TagsListField()
 
+    freeport_data = serializers.SerializerMethodField()
+
+    def get_freeport_data(self, instance):
```
mforner13, line 317: Need to add a test for this :) Have a look to see if there's something similar in tests/great_international/test_serializers.py

mforner13 (approved):

PR #1076, opened 2023-03-10
Update change log to include freeport data

LuptonM:

 - [ ] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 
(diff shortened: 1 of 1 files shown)


mforner13 (changes requested):
Release 2.9.3 has already been released and I don't think it included your change - https://github.com/uktrade/directory-cms/releases/tag/2.9.3. I think should just go in the pre-release section above until it's been released.

mforner13 (approved):

